PR #64, opened 2015-01-27
Define a Bluetooth cache, and use it in GATT navigation.

jyasskin:
This explicitly allows implementations to choose which GATT sub-procedures they
want to use to handle requests, integrates error handling into the GATT
navigation methods, and specifies that BluetoothGATT* objects are shared across
queries.

I still need to fix up:
1) the other places that specify precise GATT procedures.
2) the handling of ServiceChanged indications, which can take advantage of the explicit cache.

This is almost enough for #9: I think connection is the only imprecise algorithm left.

@marcoscaceres @scheib, let me know what you think of the demo at https://rawgit.com/jyasskin/web-bluetooth-1/information-model/index.html. I could move things around, reword, etc.

### DIFF
--- a/index.html
+++ b/index.html
@@ -684,6 +684,10 @@
           it has discovered on a device.
           The UA MAY share this cache between multiple origins accessing the same device.
           Each potential entity in the cache is either known-present, known-absent, or unknown.
+          Each known-present entity in the cache is associated with an optional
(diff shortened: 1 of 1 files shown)


jyasskin:
@jracle Are you interested in reviewing CLs like this?

scheib:
lgtm

PR #67, opened 2015-02-07
Fix #36: make BluetoothGATTCharacteristic.properties a dictionary.

### LINKED ISSUE #36: BluetoothGATTCharacteristic.properties should expose a dictionary rather than an array or bitfield
[Discussion on dev-webapi@mozilla](https://groups.google.com/d/msg/mozilla.dev.webapi/Z-ZXdtigktI/j1iwdxm6Tr4J) indicates that it would probably be more webby to expose the set of Characteristic Properties as a dictionary of boolean values rather than the bitfield that raw Bluetooth exposes. This will also let us expose the Extended Properties descriptor (3.G.3.3.3.1) uniformly with the basic properties (3.G.3.3.1.1).

The [Chrome Apps API](https://developer.chrome.com/apps/bluetoothLowEnergy#type-Characteristic) exposes a list of strings, which splits the difference. I think it'll be easier for developers to use if we expose this as a dictionary (or [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)?) rather than making them search a list.

Trying to expose the extended properties uniformly with the basic properties raises the question of whether `properties` should return a value or a `Promise`. The ATT message that discovers the Characteristic itself includes the basic properties, but if the `Extended Properties` bit is set there, the UA will need to discover a prefix of the Characteristic's Descriptors and read the value of the Characteristic Extended Properties, for at least 2 extra radio round-trips before the Characteristic can be returned. It's likely that developers won't inspect the `properties` field in most cases, since they're writing an application against a specific Service whose definition defines which characteristic properties are set, so I'm inclined to have `properties` return a `Promise`.

jyasskin:
The wording here is a little fuzzy because I don't yet have a rigorous
description of how to create a BluetoothGATTCharacteristic representing a
Characteristic, so there's no good place to describe where to read the extended
attributes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -874,7 +874,7 @@
           <dt>readonly attribute BluetoothGATTService service</dt>
           <dd>The GATT service this characteristic belongs to.</dd>
 
-          <dt>readonly attribute CharacteristicProperty[] properties</dt>
+          <dt>readonly attribute CharacteristicProperties properties</dt>
           <dd>The properties of this characteristic.</dd>
 
           <dt>readonly attribute DOMString instanceID</dt>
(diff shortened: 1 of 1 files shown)


jyasskin:
@armansito, you chimed in on the issue. Want to review the fix?

index.html
```diff
@@ -1086,6 +1086,76 @@
...
+              Returns the value of the Broadcast bit
+              of the underlying <a>characteristic properties</a>.
+            </dd>
+            <dt>readonly attribute boolean read</dt>
+            <dd>
+              Returns the value of the Read bit
+              of the underlying <a>characteristic properties</a>.
+            </dd>
+            <dt>readonly attribute boolean writeWithoutResponse</dt>
+
```
scheib: This is an odd vertical spacing between <dt> and related <dd> tags.

scheib:
lgtm

armansito:
lgtm

index.html
jyasskin: Yes. Why did I do that?

jyasskin:
Closed by 067362ffa702d223653d5b488be8098581a95719.

PR #69, opened 2015-02-10
Define when BT attributes are the same, and how to create instances of BluetoothGATT*

jyasskin:


index.html
```diff
@@ -736,28 +737,118 @@
...
-                representing the Service, Characteristic, or Descriptor, respectively,
-                and associate it with <var>entry</var>.
+                <a>create a <code>BluetoothGATTService</code>
+                representing</a> <var>entry</var>,
+                <a>create a <code>BluetoothGATTCharacteristic</code>
+                representing</a> <var>entry</var>,
+                or <a>create a <code>BluetoothGATTDescriptor</code>
+                representing</a> <var>entry</var>,
+                depending on whether <var>entry</var> is a Service, Characteristic, or Descriptor,
+                and associate the resulting <code>Promise</code> with <var>entry</var>.
```
scheib: Consider refering to the variable < var >promise< / var > instead of with < code >.

scheib: Or, is this intended to be a new promise, if so I'm initially uncertain by the existing wording. "associate the resulting Promise with entry" infers that "create a BluetoothGATT*" will have a resulting Promise. Above the cache is modified to contain Promise<BluetoothGATT*>, so perhaps this section just needs to also be updated to either find the Promise<> and or create the BluetoothGATT* and a Promose<BluetoothGATT*> for it.

If entry has no associated BluetoothGATT* instance for the current script execution environment, create a BluetoothGATTService representing entry, create a BluetoothGATTCharacteristic representing entry, or create a BluetoothGATTDescriptor representing entry, depending on whether entry is a Service, Characteristic, or Descriptor, and associate the resulting Promise with entry.

jyasskin: _promise_ refers to the `Promise` created on line 713. The `Promise` I'm associating here was created in this step. I could use wording like:

> Let _createdAttribute_ be the result of running create a ... depending on ... or Descriptor. Associate _createdAttribute_ with _entry_.

Would that read better for you?

jyasskin: The "create a `BluetoothGATT*`" algorithms that I'm adding in this PR return Promises.

index.html
```diff
@@ -736,28 +737,118 @@
...
+                the <code>Promise&lt;BluetoothGATT*></code> instance
+                associated with <var>entry</var>
                 for the current <a>script execution environment</a>.
               </li>
             </ol>
           </li>
           <li>
-            <a>Resolve</a> <var>promise</var> with <var>result</var>.
+            <a>Resolve</a> <var>promise</var> with the result of
+            <a>waiting for all</a> elements of <var>result</var>.
```
scheib: What causes an individual Promise<BluetoothGAT*> to resolve?

jyasskin: It depends on what created the Promise. In this case, if the promise was created by the "create a <code>BluetoothGATTCharacteristic" algorithm, it's resolved on line 979.

scheib: Ok, I see below where this is defined.

index.html
scheib: I overlooked the link to definition of "create a * representing", I see that this is defined and linked now.

index.html
```diff
@@ -845,6 +936,49 @@
...
+          <li>
+            <a>Get an attribute id</a> for <var>characteristic</var>,
+            and initialize <code>result.instanceID</code> to the result.
+          </li>
+          <li>
+            <a>Create a <code>CharacteristicProperties</code> instance
+            from the Characteristic</a> <var>characteristic</var>,
+            and let <var>propertiesPromise</var> be the result.
+          </li>
+          <li>Wait for <var>propertiesPromise</var> to settle.</li>
```
scheib: Reference https://heycam.github.io/webidl/#dfn-perform-steps-once-promise-is-settled

index.html
```diff
@@ -1068,13 +1202,67 @@
...
+                <tr><td><code>indicate</code></td><td>Indicate</td></tr>
+                <tr>
+                  <td><code>authenticatedSignedWrites</code></td>
+                  <td>Authenticated Signed Writes</td>
+                </tr>
+              </table>
+            </li>
+            <li>
+              If the Extended Properties bit of the <a>characteristic properties</a> is not set,
+              initialize <code><var>properties</var>.reliableWrite</code> and
```
scheib: s/properties\b/propertiesObj/.
Also, < code >< var >, is used here for an instance variable, but not below and IIRC most other places in this spec.

index.html
```diff
@@ -1068,13 +1202,67 @@
...
+                  into <var>extendedProperties</var>.
+                  Handle errors as described in <a href="#error-handling"></a>.
+                </li>
+                <li>
+                  If the previous step returned an error,
+                  <a>reject</a> <var>promise</var> with that error and abort these steps.
+                </li>
+                <li>
+                  Initialize <var>propertiesObj.reliableWrite</var> from
+                  the Reliable Write bit of <var>extendedProperties</var>.
```
scheib: Ah, < code >< var > is more widely used in these new algorithm sections ... should be used here as well.

index.html
scheib: Ah, < code >< var> is used, leave them.

index.html
scheib: lgtm as is.

scheib:
lgtm

index.html
jyasskin: I think that algorithm implies a lot more than we want here. In particular, it ends with "Return the result of calling ...", while this algorithm shouldn't return at this step. I've filed https://github.com/w3ctag/promises-guide/issues/36 to give us some terms for this, but ServiceWorker is already using "Wait for X to settle."

index.html
jyasskin: Whoops @ missing propertiesObj.

index.html
jyasskin: Yep, done.

Issue #65, opened 2015-01-29
Make connect() return a device handle


jyasskin:
#2 made `requestDevice` not imply a connection, but that implies that common uses look like:

```javascript
navigator.bluetooth.requestDevice(...)
.then(device => {device.connect().then(()=>device);})
.then(device => { /* real use */ });
```
If `connect()` returns a `BluetoothDevice` or similar, then the common use would look like:
```javascript
navigator.bluetooth.requestDevice(...)
.then(device => device.connect())
.then(device => { /* real use */ });
```
which is a bit simpler.

Issue #22, opened 2014-08-15
Forward compatibility needs?


jyasskin:
How do we need to change the spec to accommodate future extensions for non-Client/Central/GATT/LE operation?

PR #70, opened 2015-02-23
Define how devices are represented at various levels

jyasskin:
And use those definitions in the rest of the spec.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/create-devices/index.html.

### DIFF
--- a/index.html
+++ b/index.html
@@ -624,7 +624,7 @@
           <span class="issue">
             This needs a definition involving
             removing <a>BluetoothDevice</a> instances from <a>device instance map</a>s
-            and clearing out their [<span></span>[representedDevice]] fields.
+            and clearing out their [[\representedDevice]] fields.
           </span>
           For example, if the user chooses not to remember access,
           the UA might remove a device when the tab that was granted access to it is closed.
@@ -700,7 +700,7 @@
               <li>Initialize all of <var>result</var>'s optional fields to <code>null</code>.</li>
               <li>
                 Initialize an internal
-                <code><var>result</var>@[<span></span>[representedDevice]]</code> field
+                <code><var>result</var>@[[\representedDevice]]</code> field
                 to <var>device</var>.
               </li>
               <li>
@@ -711,7 +711,7 @@
                 Otherwise, initialize <code><var>result</var>.instanceId</code>
                 to the <a>device id</a> this key maps to,
                 and initialize an internal
-                <code><var>result</var>@[<span></span>[allowedServices]]</code> field
+                <code><var>result</var>@[[\allowedServices]]</code> field
                 to the <a>allowed services list</a> this key maps to.
               </li>
               <li>
@@ -796,7 +796,7 @@
               <li>
                 Initialize <code><var>result</var>.uuids</code> to
                 a <a>read only Array</a> holding the elements of <var>uuids</var>
-                that are also in <code><var>result</var>@[<span></span>[allowedServices]]</code>.
+                that are also in <code><var>result</var>@[[\allowedServices]]</code>.
               </li>
               <li>Resolve <var>promise</var> with <var>result</var>.</li>
             </ol>
@@ -877,9 +877,9 @@
               <li>
                 <a>Query the Bluetooth cache</a> for
                 the first primary GATT service
-                on <code>this@[<span></span>[representedDevice]]</code>
+                on <code>this@[[\representedDevice]]</code>
                 whose UUID is <var>service</var>
-                and whose UUID is in <code>this@[<span></span>[allowedServices]]</code>,
+                and whose UUID is in <code>this@[[\allowedServices]]</code>,
                 and let <var>promise</var> be the result.
               </li>
               <li>
@@ -895,8 +895,8 @@
               getPrimaryServices(optional BluetoothServiceUuid service)</dt>
           <dd>
             <a>Query the Bluetooth cache</a> for the primary GATT services
-            on <code>this@[<span></span>[representedDevice]]</code>
-            whose UUIDs are in <code>this@[<span></span>[allowedServices]]</code>
+            on <code>this@[[\representedDevice]]</code>
+            whose UUIDs are in <code>this@[[\allowedServices]]</code>
             and, if <var>service</var> is present, whose UUIDs are equal to <var>service</var>,
             and return the result.
           </dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -355,21 +336,21 @@
...
               <a>reject</a> <var>promise</var> with a <a>NotFoundError</a> and abort these steps.
             </li>
             <li>
-              Record the selected device in the origin's <a>allowed devices list</a>
-              and the union of the service UUIDs from <code>filters</code>
-              and <code>options.optionalServices</code>
-              in the device and origin's <a>allowed services list</a>.
+              <a title="add an allowed bluetooth device"
+                 >Add <var>device</var> to the origin's allowed devices map.</a>
+              with the union of the service UUIDs from <code>filters</code>
```
scheib: 'filters' in this section uses inconsistent < code and < var, seems that both should be used.

index.html
```diff
@@ -479,67 +476,334 @@
...
+
+        <p>
+          To <dfn>get the device id</dfn> for a <a>Bluetooth device</a> <var>device</var>,
+          the UA MUST run the following steps.
+        </p>
+        <ol>
+          <li>
+            Find the key in this origin's <a>allowed devices map</a>
+            that is the <a>same device</a> as <var>device</var>.
+            If there is no such key, this specification has a bug.
```
scheib: s/bug/error/ ... bug is too jargon for normative text?

index.html
```diff
@@ -479,67 +476,334 @@
...
+                    the value parsed for Product Version.
+                  </li>
+                </ol>
+              </li>
+              <li>
+                Initialize <code><var>result</var>.paired</code> to
+                the value of <var>device</var>'s <a>bonded flag</a>.
+              </li>
+              <li>
+                Initialize <code><var>result</var>.connected</code> to
```
scheib: Will .connected not conflict with any BR/EDR needs? Or should connected be labeled more specifically?

index.html
```diff
@@ -479,67 +476,334 @@
...
+                The UA MAY <a>populate the Bluetooth cache</a> with
+                all Services inside <var>device</var>.
+                Ignore any errors from this step.
+              </li>
+              <li>
+                If the <a>Bluetooth cache</a> contains known-present Services inside <var>device</var>,
+                add the UUIDs of those Services to <var>uuids</var>.
+              </li>
+              <li>
+                Initialize <code><var>result</var>.uuids</code> to
```
scheib: Should the result.uuids be filtered to the intersection of uuids and the origin's allowed services list for that device?  As a privacy concern if a site claims it is connecting with a heart rate monitor, but one is built into a device with many more services, I don't think we should expose to a site that those services exist without the site having asked for them (thus permitting the UA to disclose if necessary what the site is looking for).

index.html
```diff
@@ -479,67 +476,334 @@
...
+            Add a mapping from <var>device</var> to <var>promise</var>
+            in the current <a>script execution environment</a>'s <a>device instance map</a>.
+          </li>
+          <li>Return <var>promise</var>, and run the following steps <a>in parallel</a>.
+            <ol>
+              <li>Let <var>result</var> be a new instance of <a>BluetoothDevice</a>.</li>
+              <li>Initialize all of <var>result</var>'s optional fields to <code>null</code>.</li>
+              <li>
+                Initialize an internal
+                <code><var>result</var>@[<span></span>[representedDevice]]</code> field
```
scheib: Is there significance to the @[[representedDevice]] terminology? A note explaining that would help.

Why is @[[representedDevice]] needed, I haven't seen a use yet the explains the need to have it in the spec.

index.html
```diff
@@ -479,67 +476,334 @@
...
+        <p>
+          A <a>BluetoothDevice</a> instance represents a <a>Bluetooth device</a>
+          inside a particular <a>script execution environment</a>.
+        </p>
+
+        <p>
+          For each <a>script execution environment</a>,
+          the UA must maintain a <dfn>device instance map</dfn>
+          mapping <a>Bluetooth device</a>s
+          to <code>Promise&lt;<a>BluetoothDevice</a>></code> instances.
```
scheib: A map referencing Promise<BluetoothDevice> instances makes sense to me for BluetoothDevice instances that are being constructed and will later be delivered by resolving the promise. Once the promise is resolved, is it appropriate to keep the Promise<BluetoothDevice> in the map? Can the promise be vended again and resolved again?

index.html
```diff
@@ -840,9 +1119,17 @@
             Otherwise initialize <code><var>result</var>.isPrimary</code> to false.
           </li>
           <li>
-            Initialize <code><var>result</var>.device</code> from
-            the <a>BluetoothDevice</a> instance representing
-            the device in which <var>service</var> appears.
+            <a>Get the <code>BluetoothDevice</code> representing</a>
+            the device in which <var>service</var> appears,
+            and let <var>device</var> be the result.
```
scheib: < var > device is a bit misleading, because this is a Promise<BluetoothDevice>, not a BluetoothDevice.

scheib:
Review complete.

index.html
jyasskin: I'm pretty sure `.connected` is wrong in some way, either because it ignores the several kinds of connection or because `.connect()` should just return an object that represents the connection, so users wouldn't need to check `.connected`. `.connected`, if it sticks around, also needs to represent whether *this environment* is connected, not whether any environment is connected. This PR isn't aimed at fixing that, although it's nice that I've made the mistake obvious; I'll get it when fixing #65 or #22.

index.html
jyasskin: Done

index.html
jyasskin: This is gone now.

index.html
jyasskin: Done

index.html
jyasskin: It's notation from https://streams.spec.whatwg.org/#conventions, which indeed I should document. I'll add a commit doing that.

It's now used to replace "the remote device" in 2 of the better-specified algorithms. I think it'll be useful in formalizing what happens when the user revokes access to a device, too. We can null out the [[representedDevice]] and then if anything going through fails to check, it'll be an obvious mistake.

index.html
jyasskin: The Promise can be vended repeatedly, and stays resolved to the same object. I think this is the right design for this sort of thing.

index.html
jyasskin: Documented, but I can take the [[representedDevice]] internal slots back out if you think they don't add enough.

scheib: Thanks for adding the Terminology at top.  If the notation is  being used elsewhere and web spec people like it then great, it was just awkward to see it without understanding why, and the '@' in particular wasn't intuitive what it was to imply.

PR #71, opened 2015-03-03
Add advertising data to BluetoothDevice.

jyasskin:
I think this isn't part of the minimum viable release, but I'd like to give people an indication of what I think this access will eventually look like. Any ideas for how best to mark that?

@armansito @mmocny @marcoscaceres 

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/advertising-data/index.html.

### DIFF
--- a/index.html
+++ b/index.html
@@ -574,12 +574,21 @@
               <li>
                 <a>Manufacturer Specific Data</a>,
                 consisting of a map from 16-bit Company Identifier Codes to byte arrays.
+                This is treated as an empty map if it's not present.
               </li>
               <li>A <a>TX Power Level</a>, from -127dBm to 127dBm.</li>
-              <li><a>Service Data</a>, consisting of a map from UUIDs to byte arrays.</li>
+              <li>
+                <a>Service Data</a>, consisting of a map from UUIDs to byte arrays.
+                This is treated as an empty map if it's not present.
+              </li>
               <li>
                 An <a>Appearance</a>, one of the values defined by
                 the <a>org.bluetooth.characteristic.gap.appearance</a> characteristic.
+              </li>
+              <li>
+                A <a>Received Signal Strength Indication</a> (RSSI), from -127dBm to 20dBm.
+                This is received from the local Bluetooth adapter
+                rather than being sent by the remote device.
               </li>
             </ul>
           </li>
@@ -719,6 +728,11 @@
                 set <code><var>result</var>.name</code> to that string.
               </li>
               <li>
+                <a>Create a <code>BluetoothAdvertisingData</code></a> from <var>device</var>
+                and <code><var>result</var>@[[\allowedServices]]</code>,
+                and set <code><var>result</var>.adData</code> to the result.
+              </li>
+              <li>
                 If <var>device</var> has a <a>Class of Device</a>,
                 set <code><var>result</var>.deviceClass</code> to that value.
               </li>
@@ -823,6 +837,13 @@
           <dt>readonly attribute DOMString? name</dt>
           <dd>The human-readable name of the device.</dd>
 
+          <dt>readonly attribute BluetoothAdvertisingData adData</dt>
+          <dd>
+            The most recent advertising data received for this device.
+            <span class="issue">TODO: Write the algorithm to update this
+              when an advertising packet is received.</span>
+          </dd>
+
           <dt>readonly attribute unsigned long? deviceClass</dt>
           <dd>The class of the device, a bit-field defined by [[!BLUETOOTH-ASSIGNED-BASEBAND]].</dd>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -719,6 +728,11 @@
                 set <code><var>result</var>.name</code> to that string.
               </li>
               <li>
+                <a>Create a <code>BluetoothAdvertisingData</code></a> from <var>device</var>
```
scheib: The term reference "Create a BluetoothAdvertisingData" is not linking to definition correctly (perhaps the < code > tag throws off ReSpec).

This raises a spec style question as well. Unfortunately in current spec styling a term reference is not visually discernible unless clicked. Printed forms of the spec, or reading without exploratory clicks, makes it a bit hard to chase down some of these items. It would be a bit more verbose, but prose such as "Create a BluetoothAdvertisingData, using Algorithm ##,  from ..." would be much clearer.

index.html
```diff
@@ -574,13 +574,22 @@
...
+                This is treated as an empty map if it's not present.
               </li>
               <li>A <a>TX Power Level</a>, from -127dBm to 127dBm.</li>
-              <li><a>Service Data</a>, consisting of a map from UUIDs to byte arrays.</li>
+              <li>
+                <a>Service Data</a>, consisting of a map from UUIDs to byte arrays.
+                This is treated as an empty map if it's not present.
+              </li>
               <li>
                 An <a>Appearance</a>, one of the values defined by
```
scheib: Isn't this the definition of  'Appearance', as referenced below? (ditto TX Power Level, Serice Data, RSSI)

index.html
```diff
@@ -574,13 +574,22 @@
               <li>
                 <a>Manufacturer Specific Data</a>,
                 consisting of a map from 16-bit Company Identifier Codes to byte arrays.
+                This is treated as an empty map if it's not present.
               </li>
               <li>A <a>TX Power Level</a>, from -127dBm to 127dBm.</li>
-              <li><a>Service Data</a>, consisting of a map from UUIDs to byte arrays.</li>
+              <li>
+                <a>Service Data</a>, consisting of a map from UUIDs to byte arrays.
```
scheib: A reference would help terms here. Perhaps 3.C.11 (and perhaps more so that it refers to the Supplement for the types)?

index.html
```diff
@@ -907,6 +928,88 @@
...
+            All fields in <a>BluetoothAdvertisingData</a> return
+            the last value they were initialized or set to.
+          </p>
+          <dl title="[NoInterfaceObject] interface BluetoothAdvertisingData" class="idl">
+            <dt>readonly attribute unsigned short? appearance</dt>
+            <dd>
+              An <a>Appearance</a>, one of the values defined by
+              the <a>org.bluetooth.characteristic.gap.appearance</a> characteristic.
+            </dd>
+            <dt>readonly attribute short? pathLoss</dt>
```
scheib: pathLoss exposed as a convenience or for a security/privacy concern of exposing raw TX Power, RSSI? If the former, perhaps we should expose the raw data (Web exposes fundamentals when possible, then higher level primitives when needed), if the latter then the spec should explain why.

index.html
```diff
@@ -907,6 +928,88 @@
...
+              <ol>
+                <li>Let <var>mdata</var> be a new <a>Map</a> instance.</li>
+                <li>
+                  For each mapping in <var>device</var>'s <a>Manufacturer Specific Data</a>,
+                  make a new <a>read only ArrayBuffer</a> <var>bytes</var>
+                  containing the contents of its byte array,
+                  and add a mapping from the 16-bit code to <var>bytes</var> in <var>mdata</var>.
+                </li>
+                <li>
+                  Return a <a>read only Map</a> whose value is <var>mdata</var>.
```
scheib: I'm confused why term references aren't linking to their dfn-s.. but this term also isn't linking.

index.html
```diff
@@ -2609,6 +2712,25 @@
            >internal slots</a> of an object, instead of saying "the [[\y]] internal slot of x."
       </p>
 
+      <p>
+        This specification uses a few read-only types
+        that are similar to WebIDL's <a>read only Array</a>.
```
scheib: read only Array term isn't defined in this spec (maybe this is the location it should be?) and then it should probably include a reference to the WebIDL spec.

scheib:
review round complete, LGTM spec text wise, with noted suggestions. I'm confused why all the definition links aren't working for me though ;|

index.html
jyasskin: Ah, use http://rawgit.com/jyasskin/web-bluetooth-1/advertising-data/index.html instead, until https://github.com/w3c/respec/pull/409 is released.

scheib: *whew*, much better.

scheib: Not an issue (now that respec is fixed).

index.html
scheib: Not an issue (now that respec is fixed).

index.html
scheib: Not an issue (now that respec is fixed).

index.html
scheib: Not an issue (now that respec is fixed).

index.html
scheib: Not an issue (now that respec is fixed).

scheib:
All comments are fixed by the respec fix with the exception of the pathloss comment: https://github.com/WebBluetoothCG/web-bluetooth/pull/71#discussion_r26340073

index.html
jyasskin: It's convenience, since nobody's presented a use case for using either TX Power or RSSI on their own. I think the closest anyone's come is suggesting that TX Power can be a proxy for battery status, but that only works if you know the exact device, and the connection you need for that may as well fetch the battery service.

I see the point about preferring to expose fundamentals, but if we do that, we're also going to need to document that just about every use needs to subtract the fields. Do you think we should expose all three?

scheib: Given that we don't see a security, privacy, implementation, cross-platform, or other need for the abstraction I would recommend exposing the primitives only. It's quite nice to mention in the spec how they can be used by subtracting them, but also not required I think.

jyasskin: Done. I still disagree, but we can discuss on an issue.

scheib:
LGTM

PR #72, opened 2015-03-03
Fix #61: Be clear that the UA must show a prompt even if no devices were found.

### LINKED ISSUE #61: Describe what happens when requestDevice's scan finds no devices
We need the UA to show the dialog anyway so that users have a chance to make the relevant device discoverable, and so that sites can't use the timing of the result to identify devices a user does _not_ have.

jyasskin:
Demo at [https://rawgit.com/jyasskin/web-bluetooth-1/empty-scan-result/index.html](https://rawgit.com/jyasskin/web-bluetooth-1/empty-scan-result/index.html#widl-BluetoothDiscovery-requestDevice-Promise-BluetoothDevice--sequence-BluetoothScanFilter--filters-RequestDeviceOptions-options)

### DIFF
--- a/index.html
+++ b/index.html
@@ -322,14 +322,17 @@
             <li>
               <a>Scan for devices</a> with
               the union of all <code>services</code> sequences in <code><var>filters</var></code>
-              as the <var>set of <a>Service</a> UUIDs</var>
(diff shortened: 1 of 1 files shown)


scheib:
LGTM

PR #84, opened 2015-03-26
Add more content to the draft Web Bluetooth CG Charter:

jyasskin:
* List stream communication as out-of-scope.
* Suggest that the Spec be dual-licensed.
* Mention the security model, test suite, and polyfills as extra deliverables.
* Rephrase the WebAppSec liason, since they should review more than just our use of "secure contexts".
* Mention possibly asking for changes to Bluetooth itself, through the SIG.
* Mention that contributions can come in through GitHub.

@anssiko

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -59,6 +59,7 @@ This software may include:
 * A test suite for the Web Bluetooth API.
 * Polyfills for the Web Bluetooth on top of existing native APIs
   like [Chrome Apps](https://github.com/WebBluetoothCG/chrome-app-polyfill) and Cordova.
+* Demo applications and debugging tools for developers using the Web Bluetooth API.
 
 ## Dependencies or Liaisons
 
--- a/charter.md
+++ b/charter.md
@@ -24,7 +24,8 @@ that implement Bluetooth 4.0, 4.1, or 4.2.
 
 ### Out of Scope
 
-{TBD: Identify topics known in advance to be out of scope}
+* Communication over arbitrary L2CAP channels, RFCOMM channels,
+  or other stream and socket-based channels.
 
 ## Deliverables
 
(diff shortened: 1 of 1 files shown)


charter.md
```diff
@@ -34,15 +35,37 @@ that implement Bluetooth 4.0, 4.1, or 4.2.
 standard using the Generic Attribute Profile (GATT). See the
 [Web Bluetooth API](http://webbluetoothcg.github.io/web-bluetooth/) that serves as the
 initial input.
+This specification is a candidate for the Dual License as specified in
```
scheib: I don't object to this line, but am curious what the source of its selection is. Spot checking a few e.g. WebApps documents I don't see a reference to this document. I see some related copyright and attribution comments in https://www.w3.org/community/about/agreements/cla/. It would be good to document at least in change history the source of this.

charter.md
```diff
@@ -34,15 +35,37 @@ that implement Bluetooth 4.0, 4.1, or 4.2.
...
+
+* A description of the Web Bluetooth API's security model,
+  including one or more potential ceremonies through which a user grants limited access to a device.
+
+### Test Suites and Other Software
+
+The group MAY produce test suites and other software to support the Specifications.
+This software will be hosted in the [WebBluetoothCG GitHub organization](https://github.com/WebBluetoothCG/)
+under the [Apache 2](http://www.apache.org/licenses/LICENSE-2.0.html) license.
+This software may include:
```
scheib: Might as well include the test peripheral app?

scheib:
LGTM

charter.md
jyasskin: I copied this line from http://www.w3.org/2014/06/webapps-charter.html#url. The idea is that it would let a group like WHATWG fork the spec if the W3C ever freezes again.

anssiko: Looks good to me. I suggest we merge this and loop the revised version of the Charter via Wayne in the mail thread.

charter.md
anssiko: I'm fine with these additional deliverables.

jyasskin: Added "Demo applications and debugging tools for developers using the Web Bluetooth API." to cover the test peripheral app.

charter.md
wcarr: WebApps WG has that in their charter because W3C Working Groups publish under the W3C Document License which doesn't allow forking for use in specs.  There's an experiment that allows HTML WG extensions and this one Web Apps WG spec dual license. But, that is only relevant for specs in S3C Working Groups. Community Groups already publish under the CLA which allows forking, so it's pointless for a Community Group spec. <a href="https://www.w3.org/community/about/agreements/cla/#id2">https://www.w3.org/community/about/agreements/cla/#id2</a>

PR #92, opened 2015-03-30
Fix assertion that Peripherals can only receive one connection.

jyasskin:
@g-ortuno @armansito 

I suspect we should rewrite the Introduction to be more explanatory, but this'll fix the immediate issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -126,9 +126,9 @@
         BTLE defines multiple roles that devices can play.
         The <dfn>Broadcaster</dfn> and <dfn>Observer</dfn> roles are
         for transmitter- and receiver-only applications, respectively.
-        The <dfn>Peripheral</dfn> role is able to receive a single connection.
-        The <dfn>Central</dfn> role supports multiple connections,
-        and is responsible for creating any connections to <a>Peripheral</a> devices.
+        Devices acting in the <dfn>Peripheral</dfn> role can receive connections,
+        and devices acting in the <dfn>Central</dfn> role can connect to <a>Peripheral</a> devices.
+      </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -126,9 +126,9 @@
         BTLE defines multiple roles that devices can play.
         The <dfn>Broadcaster</dfn> and <dfn>Observer</dfn> roles are
         for transmitter- and receiver-only applications, respectively.
-        The <dfn>Peripheral</dfn> role is able to receive a single connection.
-        The <dfn>Central</dfn> role supports multiple connections,
-        and is responsible for creating any connections to <a>Peripheral</a> devices.
+        Devices acting in the <dfn>Peripheral</dfn> role can receive connections,
```
g-ortuno: What about "can receive multiple connections"?

jyasskin: That depends on the controller, right? Not all Peripherals will be able to accept multiple connections, especially since we're hoping to support BT4.0 devices with this spec.

g-ortuno: Isn't the meaning of `can receive connections` the same as `can receive multiple connections`, except the latter is clearer? If BT4.0 is a concern what about: `can receive one or, if supported, multiple connections`. Either way, I'm definitely nitpicking here, if the intro is gonna be rewritten then this LGTM.

jyasskin: For the introduction, I'm not really going for precision, just the general idea to help people navigate the rest of the spec. Even the full list of possible roles might be too much.

pzboyz: You could try and start to use the terms Master and Slave, which device started the connection is not hugely important and it is highly likely that all devices will soon have a GATT Database, so the usefulness of the terms Peripheral and Central terms become less useful and at worst are misleading.

PR #95, opened 2015-03-31
[charter] simplify Contribution Mechanics

wcarr:
As discussed on [CG mail list] (https://lists.w3.org/Archives/Public/public-web-bluetooth/2015Mar/0024.html), simplify the contribution rules.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -145,7 +145,7 @@ and does not unreasonably favor or discriminate against any group participant or
 
 The group will conduct all of its technical work in public.
 
-Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/). Other discussions take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/). For convenience, GitHub issues and pull requests will be archived to the group's [public-web-bluetooth-contrib/](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/) list. Other contributions MAY be directly posted to the contrib list.
+Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/). Other discussions MAY take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/). Other contributions MAY be also be directly posted to the contrib list.  For convenience, GitHub issues and pull requests will be archived to the group's [public-web-bluetooth-contrib/](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/) list. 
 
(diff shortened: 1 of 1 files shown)


charter.md
```diff
@@ -96,24 +96,11 @@ The CLA applies to these Community Group Reports.
...
-["contrib" list](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/)
-or [group list](http://lists.w3.org/Archives/Public/public-web-bluetooth/),
-with subject line starting "[short-name-for-spec]PR"
-or by making the request in a GitHub issue.
-The pull request that results from this must indicate who the request was made for
-and provide the URL for the archived request.
-For software, the GitHub CONTRIBUTING and LICENSE files describe how to contribute
-and under what license Contributions are made.
-
-GitHub issues and pull requests will be archived to the
```
jyasskin: Double-checking that you meant to remove this paragraph about where GitHub things are archived to. There's a requirement somewhere on the W3C site that there always needs to be an archive on their servers, but that requirement didn't say it needs to be listed in the charter.

jyasskin:
Other than my one question, this looks good. Thanks!

charter.md
wcarr: I forgot you aren't using a W3C's GitHub (e.g. github.com/w3c/web-nfc).  W3C backs that up - so that fulfills the archiving requirement.  I think they back it up so in the event GitHub had an accident or disappeared, they'd still have it.  But, that line would be good to keep that line in even if you switch to using W3C's GitHub.  Also, you could send it to the group's public-web-bluetooth-contrib list - an important purpose of archiving the discussions is to be able to track contributions. It could say: "GitHub issues and pull requests will be archived to the [public-web-bluetooth-contrib](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/) list."  That would make it easy for people to find it on the CG page - it's right with the general public list.

jyasskin: Right. @dontcallmedom picked the -log list in https://lists.w3.org/Archives/Public/public-web-bluetooth/2015Mar/0006.html. I don't have a preference for which list, but it'd be good to have the W3C write down a policy for this.

jyasskin: For now, maybe just avoid changing the existing line in this PR, and we'll take the rest of your change, and change the list if necessary in a separate PR.

wcarr: I'll do a different PR to change the transparency section to add update it for Github - adding the links to the log there along with the links for the mail list, contrib list.  So that line would go but it would reappear in the transparency section along with the other lists.

jyasskin:
'k, I'll take this with the open question about which list to mirror Github to, and we'll either fix the wording in another PR or have Dominique change where the mirroring happens.

PR #98, opened 2015-04-06
Update the archive address for github issues, in the charter.

jyasskin:
https://lists.w3.org/Archives/Public/public-web-bluetooth/2015Apr/0001.html confirmed the preference to put the archive on a list folks aren't automatically subscribed to, to avoid spam. I'll make sure this list is mentioned in the Mailing Lists on https://www.w3.org/community/web-bluetooth/.

@marcoscaceres @wcarr

[Review on Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/98)

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -145,7 +145,11 @@ and does not unreasonably favor or discriminate against any group participant or
 
 The group will conduct all of its technical work in public.
 
-Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/). Other discussions MAY take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/). Other contributions MAY be also be directly posted to the contrib list.  For convenience, GitHub issues and pull requests will be archived to the group's [public-web-bluetooth-contrib/](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/) list. 
+Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/).
+Other discussions MAY take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/).
+Other contributions MAY be also be directly posted to the contrib list.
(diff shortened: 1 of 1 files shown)


charter.md
```diff
@@ -145,7 +145,11 @@ and does not unreasonably favor or discriminate against any group participant or
 
 The group will conduct all of its technical work in public.
 
-Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/). Other discussions MAY take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/). Other contributions MAY be also be directly posted to the contrib list.  For convenience, GitHub issues and pull requests will be archived to the group's [public-web-bluetooth-contrib/](http://lists.w3.org/Archives/Public/public-web-bluetooth-contrib/) list. 
+Discussions MAY take place on the group's [public mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/).
+Other discussions MAY take place using Issues in the group's [GitHub repository](https://github.com/WebBluetoothCG/).
+Other contributions MAY be also be directly posted to the contrib list.
```
marcoscaceres: Maybe link to contrib list?

marcoscaceres:
LGTM.

charter.md
wcarr: The contrib list link is in the transparency section too.

One of the things that the discussions here brought up was the charter template that this came from should be clearer that most of the process rules are just for emergencies, when the usual ways of doing things don't work.  I changed the template to make that clearer.  I'll do a PR to add those clarifications.

jyasskin: +1 on linking all mentions. That wasn't the goal of this PR, but I'll go through and add links in another commit.

@wcarr: I'd like not to bog down wordsmithing this charter. I agree that the wording you've applied to the NFC charter is a bit clearer about the decision procedure, but I don't think the wording here is _bad_, so I'd like to move this through the adoption process without demanding that it keep up with iteration on the new template.

PR #99, opened 2015-04-07
updates for consistency with template

wcarr:
broke some long paragraphs up, put in headers and made clear that the CG can do things however they like and the formal process in the charter is only if there are objections with the informal group process.  and that should be rare.

it's find to drop this as it's late.  Most of what looks like deletions are just moving to a separate paragraph with headings.

[Review on Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/99)

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -1,8 +1,8 @@
 # Web Bluetooth Community Group Charter
 
-* This Charter: TBD
+* This Charter: https://github.com/WebBluetoothCG/web-bluetooth/blob/charter-2015-04/charter.md
 * Previous Charter: None
-* Start Date: 2015-??-??
+* Start Date: 2015-05-07
 
 ## Goals
 
--- a/charter.md
+++ b/charter.md
@@ -150,7 +150,7 @@ When the group discusses an issue on the [mailing list](http://lists.w3.org/Arch
 and there is a call from the group for assessing consensus,
 after due consideration of different opinions, the Chair should record a decision and any objections.
 
-A common way to determine consensus for important decisions is to conduct a Call for Consenus (CfC) where the Chair puts a proposal to the group on the public mail list and asks for feedback from the Participants within some period of time that is at least 7 days. Not responding indicates going along with the proposal. Direct feedback is encouraged, especially to weigh the degree of consensus when there is dissent.
+A common way to determine consensus for important decisions is to conduct a Call for Consensus (CfC) where the Chair puts a proposal to the group on the public mail list and asks for feedback from the Participants within some period of time that is at least 7 days. Not responding indicates going along with the proposal. Direct feedback is encouraged, especially to weigh the degree of consensus when there is dissent.
 
 Participants may call for an online vote if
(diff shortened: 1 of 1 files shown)


jyasskin:
I think this is all just wordsmithing, so I'd rather not spend the time iterating on it, even though I don't have any objections to the content. I expect we can explain in emails that the informal processes are preferred, instead of relying on the charter to explain it.

@marcoscaceres or others with more W3C experience than I have, let me know if you disagree and want to take this.

---



---

<sup>**[charter.md, line 100 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/99#-JmGPdPjQm0GjC4oAi1_)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/712f6b9f239d3955e9b367dde12aa57a1016f2e8/charter.md#L100)):</sup>
This is something the editors will have to pay attention to when accepting changes, but the [CONTRIBUTING](https://github.com/WebBluetoothCG/web-bluetooth/blob/charter/CONTRIBUTING.md) file is clear about it too, so having it in the charter too probably isn't necessary.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/99)

charter.md
```diff
@@ -139,8 +149,13 @@ This group will seek to make decisions when there is consensus.
 When the group discusses an issue on the [mailing list](http://lists.w3.org/Archives/Public/public-web-bluetooth/)
 and there is a call from the group for assessing consensus,
 after due consideration of different opinions, the Chair should record a decision and any objections.
+
+A common way to determine consensus for important decisions is to conduct a Call for Consenus (CfC) where the Chair puts a proposal to the group on the public mail list and asks for feedback from the Participants within some period of time that is at least 7 days. Not responding indicates going along with the proposal. Direct feedback is encouraged, especially to weigh the degree of consensus when there is dissent.
```
marcoscaceres: s/Consenus/Consensus

wcarr:
The main justification for accepting it is that it is changes to make things that weren't understood here clearer, mostly just with some headers and paragraph breaks.

The main reason for accepting the changes is that CGs often use other CGs Charter as their template (like this one used the web nfc charter as the starting point), rather than using the actual template, so it's nice to have them consistent as of when they go for approval.  So, while not essential, it would be nice for this to match the template now (and the web nfc charter) for the process part.

jyasskin:
FWIW, this one used https://www.w3.org/community/council/wiki/Templates/CG_Charter, the first result for https://www.google.com/search?q=cg+charter+template, as its template. If that's not "the" template, it should link there.

wcarr:
That one hasn't been updated for CGs that use GitHub, like it says that all Contributions have to be to the public mail list or contrib mail list.  It doesn't have GitHub as a way to Contribute.  Thanks for mentioning it.  I'd forgotten it wasn't updated.

PR #102, opened 2015-04-11
Switch to ReSpec's contiguous IDL mode.

jyasskin:
Demo: https://rawgit.com/jyasskin/web-bluetooth-1/contiguous-idl/index.html

https://github.com/jyasskin/web-bluetooth-1/compare/contiguous-idl...jyasskin:wording-cleanups has a few cleanups applied. I still need to fix a couple bugs in ReSpec:

* Fix Contiguous IDL mode to link the UUID typedef.
* Figure out why definitions in notes aren't linking.

And a couple bugs in the spec that I noticed when I went over the whole text:
* writeValue needs a definition.
* Descriptor.readValue and writeValue need definitions.
* BluetoothInteraction needs promise-based methods, a store the UA looks up the IDs in, and a description of what happens when the ID isn't present.
* Say something about how `ieee_11073-20601_regulatory_certification_data_list` is mapped to an attribute name.

[Review on Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/102)

### DIFF
--- a/index.html
+++ b/index.html
@@ -317,17 +317,43 @@
             optional RequestDeviceOptions options);
         };
       </pre>
-      <p>
-        <dfn for="BluetoothScanFilter" title="services">BluetoothScanFilter.services</dfn>
-        holds a list of <a>Service</a> UUIDs that a device must support
-        to match the <a>BluetoothScanFilter</a>.
-      </dl>
+      <div class="note" title="BluetoothScanFilter attributes" dfn-for="BluetoothScanFilter">
+        <p>
+          <dfn>services</dfn> holds a list of <a>Service</a> UUIDs that a device must support
+          to <a title="matches a filter">match</a> the <a>BluetoothScanFilter</a>.
+        </p>
+      </div>
 
-      <p dfn-for="RequestDeviceOptions">
-        In <a>RequestDeviceOptions</a>, <dfn>optionalServices</dfn> holds
-        a list of <a>Service</a> UUIDs that aren't required for the website to use a device,
-        but that the website can take advantage of if they're present.
+      <p>
+        A device <dfn>matches a filter</dfn> <var>filter</var> if
       </p>
+      <ul>
+        <li><code><var>filter</var>.services</code> is not empty and</li>
+        <li>
+          the UA has received advertising data, an <a>extended inquiry response</a>,
+          or a service discovery response indicating that
+          the device supports each of the <a>Service</a> UUIDs
+          included in <code><var>filter</var>.services</code> as a primary (vs included) service.
+        </li>
+      </ul>
+
+      <p class="note">
+        The list of Service UUIDs that a device advertises
+        might not include all the UUIDs the device supports.
+        The advertising data does specify whether this list is complete.
+        If a website filters for a UUID that a nearby device supports but doesn't advertise,
+        that device might not be included in the list of devices presented to the user.
+        The UA would need to connect to the device to discover the full list of supported services,
+        which can impair radio performance and cause delays, so this spec doesn't require it.
+      </p>
+
+      <div class="note" title="RequestDeviceOptions attributes" dfn-for="RequestDeviceOptions">
+        <p>
+          <dfn>optionalServices</dfn> holds
+          a list of <a>Service</a> UUIDs that aren't required for the website to use a device,
+          but that the website can take advantage of if they're present.
+        </p>
+      </div>
 
       <p>
         When <code><dfn for="BluetoothDiscovery">requestDevice</dfn>(<var>filters</var>, <var>options</var>)</code> is invoked,
@@ -439,26 +465,6 @@
           </p>
         </aside>
       </p>
-
-      <p>
-        A device <dfn>matches a filter</dfn> <var>filter</var> if
-      <ul>
-        <li><code><var>filter</var>.services</code> is not empty and</li>
-        <li>
-          the UA has received advertising data, an <a>extended inquiry response</a>,
-          or a service discovery response indicating that
-          the device supports each of the <a>Service</a> UUIDs
-          included in <code><var>filter</var>.services</code> as a primary (vs included) service.
-      </ul>
-
-      <p class="note">
-        The list of Service UUIDs that a device advertises
-        might not include all the UUIDs the device supports.
-        The advertising data does specify whether this list is complete.
-        If a website filters for a UUID that a nearby device supports but doesn't advertise,
-        that device might not be included in the list of devices presented to the user.
-        The UA would need to connect to the device to discover the full list of supported services,
-        which can impair radio performance and cause delays, so this spec doesn't require it.
 
       <p>
         To <dfn>scan for devices</dfn> with
--- a/index.html
+++ b/index.html
@@ -912,7 +912,7 @@
 
         <div class="note" title="BluetoothDevice attributes">
           <p>
-            <dfn>instanceID</dfn> returns the opaque identifier assigned to this device.
+            <dfn>instanceID</dfn> is the opaque identifier assigned to this device.
             This identifier MUST uniquely identify a device to the extent that
             the UA can determine that two Bluetooth connections are to the same device.
             For example, a paired device MUST keep the same <code>instanceID</code>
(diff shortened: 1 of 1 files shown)


scheib:
Started reviewing now.

This patch is too big. github doesn't display, reviewable bogs down considerably and froze on my at one point.

scheib:
<sup>**[index.html, line 323 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuAI9RYJZizOk6E8fe)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L323)):</sup>
Unsure if you'd like to do content editing as well with this refactor. If so, probably pull the "matches a filter" definition up to this section. Or, at least link 'match' here with `<a title="matches a filter">match</a>`

---

<sup>**[index.html, line 327 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuGB0Fw8f05-mcq2dB)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L327)):</sup>
Drop the `In `. There's lack of consistency with the dictionaries BluetoothScanFilter and RequestDeviceOptions, and I miss the clarity of the [old version](http://webbluetoothcg.github.io/web-bluetooth/#widl-BluetoothScanFilter-services) in sections with headers. They are small interfaces, though, so the single lines are fine. I prefer the directness of `BluetoothScanFilter.services holds ...`

---

<sup>**[index.html, line 907 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuOJXiqd8U1eDK6O1r)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L907)):</sup>
A note here seems odd, these are the attribute definitions and seem appropriate to be in 'normal' spec prose. Perhaps sectioned off with a header.

---

<sup>**[index.html, line 981 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuQ1zXyPDpiq96G0Hx)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L981)):</sup>
Consider leading with the term being defined. `connectGATT() is used to obtain a  BluetoothGATTRemoteServer from a device. The UA MUST return ...`

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

scheib:
(review still underway, breaking it up.)

jyasskin:
<sup>**[index.html, line 323 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuAI9RYJZizOk6E8fe)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L323)):</sup>
And you were just complaining that the patch is too big. ;) Done with the `<a>`, although not uploaded yet. I wonder if these should be Notes for the same reason I'm using the notes below.

---

<sup>**[index.html, line 327 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuGB0Fw8f05-mcq2dB)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L327)):</sup>
I can do sections with headers in this version too. I actually have more control over the titles of the sections, too.

However, the paragraphs that just name fields is more consistent with other specs: https://fetch.spec.whatwg.org/#dom-request-headers

Actually, the Fetch spec doesn't even describe the fields of its dictionaries, which lends more credence to the idea that they should be notes at most.

---

<sup>**[index.html, line 907 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuOJXiqd8U1eDK6O1r)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L907)):</sup>
They're not normative, in that they don't say "name MUST return Xyz". Instead, attributes MUST return the last value they're set to, and the rest of the spec talks about setting them where necessary. I haven't yet added the "MUST return the last value they're set to" text for most of them, and instanceID in particular is an outlier in that it has way too much MUST that's already covered by how it's set.

This is somewhat inconsistent with other specs, but I found one that does it this way ... that I'm trying to re-find so I can imitate it in the "MUST return the last value they're set to" text. I think it would be difficult to invert this and give user-friendly descriptions of the attributes that are also normative.

---

<sup>**[index.html, line 981 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuQ1zXyPDpiq96G0Hx)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L981)):</sup>
https://fetch.spec.whatwg.org/#dom-response-redirect uses the awkward construction "The static redirect(url, status) method, when invoked, must run these steps:". Everything in that spec starts with "The", and talks about the method doing things instead of the UA. http://www.w3.org/TR/service-workers/#navigator-service-worker-getRegistration uses the same construct, but skips the "The".

@marcoscaceres, any preferences?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

scheib:
Reviewed files:
- index.html @ r2

---

<sup>**[index.html, line 930 \[r2\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmyAaSrxlBjmJX-MUSc)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/1bf2faf58ad22a14c544c01bb224dbd3bad454aa/index.html#L930)):</sup>
'provides' or 'is' instead of 'returns'?

---

<sup>**[index.html, line 1489 \[r2\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmyBngxlK87umnYWtF5)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/1bf2faf58ad22a14c544c01bb224dbd3bad454aa/index.html#L1489)):</sup>
For another patch: UUID.parse isn't  specified.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

scheib:
Done looking over everything -- LGTM

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

marcoscaceres:
> @marcoscaceres, any preferences?
I don't have any strong preference (though I think I prefer Anne's style) - whatever you think makes the spec more legible :)

anssiko:
@jyasskin You can ping me if you have issues with the `noLegacyStyle` flag. I implemented the feature in ReSpec, initially scratching one's own itch, so the migration has some quirks. I tried to update the docs to give some guidance, but likely did not capture everything: http://www.w3.org/respec/ref.html#nolegacystyle

jyasskin:
@anssiko Thanks. The contiguous IDL support actually came out of a bug I filed about `noLegacyStyle`dropping the member documentation (https://github.com/w3c/respec/issues/334). I'm using `noLegacyStyle` in this patch because of https://github.com/w3c/respec/issues/427, but once I fix that in ReSpec, I can probably drop `noLegacyStyle` entirely.

---





<sup>**[index.html, line 930 \[r2\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmyAaSrxlBjmJX-MUSc)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/1bf2faf58ad22a14c544c01bb224dbd3bad454aa/index.html#L930)):</sup>
Yeah, in a note we should use simpler English. Done.

---

<sup>**[index.html, line 987 \[r1\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmuQ1zXyPDpiq96G0Hx)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/56d3238dea4fcfe4475af1e45c7ad7b1ba2a097b/index.html#L987)):</sup>
@marcoscaceres Thanks, I've switched to Anne's style from the fetch spec.

@scheib, I think you have a good point that it'd be good to have English descriptions of the purpose of these methods in addition to the standardese describing how they work. I'm thinking about moving the "Foo attributes" notes to always be right after their IDL, and then adding prose descriptions of the methods there too. I could put examples immediately after that...

---

<sup>**[index.html, line 1489 \[r2\]](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102#-JmyBngxlK87umnYWtF5)** ([raw file](https://github.com/webbluetoothcg/web-bluetooth/blob/1bf2faf58ad22a14c544c01bb224dbd3bad454aa/index.html#L1489)):</sup>
`UUID.parse` doesn't exist anymore, since UUIDs are now just `DOMString`s. I've removed the extraneous calls.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

scheib:
Reviewed files:
- index.html @ r3

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/102)

Issue #60, opened 2014-12-18
NavigatorBluetooth can be a partial interface


foolip:
https://webbluetoothcg.github.io/web-bluetooth/#interface-wiring

The style of the spec in this section is a bit atypical. A section title "Extensions to the Navigator Interface" and a partial interface Navigator would be more in line with other specs:

http://dev.w3.org/csswg/cssom/#extensions-to-the-document-interface
http://dev.w3.org/csswg/cssom-view/#extensions-to-the-element-interface
http://w3c.github.io/webcomponents/spec/shadow/#extensions-to-element-interface

jyasskin:
@g-ortuno pointed out that Chrome's implementation already uses partial interfaces here: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/bluetooth/NavigatorBluetooth.idl, so I'm going to change both this and {`BluetoothDiscovery`,`BluetoothInteraction`} to partial interfaces.

PR #108, opened 2015-04-24
Use partial interfaces to replace some `implements` statements.

jyasskin:
https://rawgit.com/jyasskin/web-bluetooth-1/partial-interfaces/index.html

@g-ortuno Does this fit in well with your implementation work?

[Review on Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/108)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1890,7 +1890,7 @@ return navigator.bluetooth.requestDevice({
           For each known GATT <a>Characteristic</a>, the UA MUST maintain
           an <dfn>active notification context set</dfn> of <a>BluetoothInteraction</a> objects.
           This is a single set for the whole UA,
-          pointing to the <a href="#idl-def-navigatorbluetooth-bluetooth"><code>navigator.bluetooth</code></a> object
+          pointing to the <a href="#idl-def-navigator-bluetooth"><code>navigator.bluetooth</code></a> object
           for each separate <a>script execution environment</a> that has registered for notifications.
         </p>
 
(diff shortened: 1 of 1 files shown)


scheib:
LGTM, especially fixing #60 is cleaner.

---

Reviewed files:
- index.html @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/webbluetoothcg/web-bluetooth/108)

g-ortuno:
Implementing a partial interface seems to be more complex that originally thought. We are currently exploring an option to implement the spec as it currently stands.

scheib:
@g-ortuno and I have discussed implementation a bit further, the partial interfaces do have some overhead, perhaps more than just using C++ composition to solve the previous multiple implements version. @g-ortuno will give composition code a shot with multiple implements as that may actually be cleaner, and if so perhaps we shouldn't do this split.

jyasskin:
@g-ortuno @scheib 5e368f09a42aed8660b59a0a9e0e93ded7c4e0fe has the remaining change to extend Navigator with a partial interface instead of using `implements` (now that `BluetoothInteraction` is gone). Let me know if you want this, or if I should just close it.

scheib:
The 'implements' method is simplest to implement in Blink, though both are possible. Choose what feels best for the spec, but if it's a tie stay with 'implements' for impl sake.

scheib:
Sorry, with @g-ortuno reminding me to look closer, Navigator with a partial interface is more consistent in Blink. It is more work for interfaces beyond Navigator to also support being supplemented with partial interfaces, so all other interfaces we'd prefer implements.

jyasskin:
Sounds good. This change now makes only Navigator use a partial interface, with no partial chunks on other interfaces.

PR #111, opened 2015-05-15
Specify writeValue on characteristics and both readValue and writeValue on descriptors.

### LINKED ISSUE #26: Specify how writeValue() works [was: Support ReliableWrite]
Do we need to have BluetoothGATTBeginReliableWrite,  BluetoothGATTEndReliableWrite,  BluetoothGATTAbortReliableWrite, or did I miss something?

### LINKED ISSUE #44: Enforce a maximum length on writes
BT4.1 3.F.3.2.9 Long Attribute Values says:

> The maximum length of an attribute value shall be 512 octets.

We should fail an attempt to write a value longer than that before sending it down to the OS or over the radio, in order to protect any fixed-length buffers hiding in those layers.

jyasskin:
Demo at https://rawgit.com/jyasskin/web-bluetooth-1/specify-read-write/index.html.

I also switched ArrayBuffer arguments to BufferSource per advice in
https://heycam.github.io/webidl/#idl-buffer-source-types

Fixes #26, postponing the options dictionary to the second version of the API.
Also fixes #44.

@armansito @shawnjohnjr see if this is precise enough. We may also want a new bug for adding reliable-write/transaction support in the next version.

[Review on Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/111)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1585,7 +1585,7 @@ return navigator.bluetooth.requestDevice({
             Promise&lt;sequence&lt;BluetoothGATTDescriptor>>
               getDescriptors(optional BluetoothDescriptorUUID descriptor);
             Promise&lt;ArrayBuffer> readValue();
-            Promise&lt;void> writeValue(ArrayBuffer value);
+            Promise&lt;void> writeValue(BufferSource value);
             Promise&lt;void> startNotifications();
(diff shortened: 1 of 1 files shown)


scheib:
LGTM

PR #113, opened 2015-05-19
Remove attribute instance IDs, and merge BluetoothDiscovery and BluetoothInteraction.

### LINKED ISSUE #51: Remove Instance IDs
The `instanceid` attribute currently exists on `BluetoothDevice`, `BluetoothGATTService`, `BluetoothGATTCharacteristic`, and `BluetoothGATTDescriptor`. The `instanceid` can be used to 

1. Disambiguate multiple instances of objects with the same UUID. 
2. Get an object for the instance from the navigator.bluetooth.get* methods (`BluetoothInteraction` interface). 

However, instance IDs may not be exposed on all platform APIs. E.G. See Apple's [CBService](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBService_Class/index.html#//apple_ref/occ/instp/CBService/characteristics).

Implementations of Web Bluetooth could manufacture IDs even when the OS API does not provide them to work around this. IDs would need to be invalidated whenever the association can't be maintained, e.g. device disconnection or on Service Change.

Further, the information does not seem essential to a minimum viable Web Bluetooth specification. Multiple service, characteristic, descriptor instances will be returned as unique objects, and applications can assign ID if necessary to them in the same way this specification wound need to on operating systems that do not expose an instance ID.

### LINKED ISSUE #52: Remove BluetoothInteraction get* Methods
https://webbluetoothcg.github.io/web-bluetooth/#idl-def-BluetoothInteraction currently offers
```
    Promise<BluetoothGATTService>        getService (DOMString serviceInstanceId);
    Promise<BluetoothGATTCharacteristic> getCharacteristic (DOMString characteristicInstanceId);
    Promise<BluetoothGATTDescriptor>     getDescriptor (DOMString descriptorInstanceId);
```

I motion we remove these from the initial spec
a. To keep initial spec as simple as possible. Access to Service, Characteristic, and Descriptor objects is already exposed by the chain of objects, get*, and returned promises.
b. Scanning e.g. Android SDK, I don't see a way for the implementation to efficiently implement loading Service, Characteristic, and Descriptor objects from ID, the implementation would need to enumerate the full hierarchy anyway to search for matches.
c. InstanceIDs may not be easily accessible on all platforms, see #51 which proposes removing all `instanceid` attributes.

jyasskin:
The Mac/iOS platforms don't expose a persistent instance ID for attributes. I
believe they use object identity within a process to distinguish multiple
attributes with the same UUID, although that's not documented. This patch also
documents that the web platform will use object identity for this purpose.

Removing instance IDs makes BluetoothInteraction empty, so I don't need
Bluetooth to have multiple components anymore.

Fixes #51 and fixes #52.

### DIFF
--- a/index.html
+++ b/index.html
@@ -165,7 +165,7 @@
 
         <p>
           When a website requests access to devices using
-          <code><a for="BluetoothDiscovery">requestDevice</a></code>,
+          <code><a for="Bluetooth">requestDevice</a></code>,
           it gets the ability to access all GATT services mentioned in the call.
           The UA MUST inform the user what capabilities these services give the website
           before asking which devices to entrust to it.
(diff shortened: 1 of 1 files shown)


scheib:
LGTM

PR #121, opened 2015-05-27
Add § links to the current section.

### LINKED ISSUE #120: [Feature Request] Link button in IDL definitions
Is there anyway we can add link buttons to IDL definitions to get the url to the IDL faster? Every time I want to reference the IDL I'm looking at, I have to open the inspector copy the id and paste it into the address bar.

jyasskin:
@g-ortuno, does this fix #120?

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/permalinks/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,6 +9,9 @@
       var respecConfig = {
           specStatus: "CG-DRAFT",
           shortName:  "web-bluetooth",
+
+          includePermalinks: true,
+
           editors: [
             { name: "See contributors on GitHub",
               url: "https://github.com/WebBluetoothCG/web-bluetooth/graphs/contributors"


g-ortuno:
Some of the interface names are clickable and some are not. For example you can click on [BluetoothGATTDescriptor](https://rawgit.com/jyasskin/web-bluetooth-1/permalinks/index.html#dom-bluetoothgattdescriptor) but not on [BluetoothUUID](https://rawgit.com/jyasskin/web-bluetooth-1/permalinks/index.html#idl-def-bluetoothuuid)

jyasskin:
Right, names in the IDL are clickable when there's a definition for the name somewhere outside of the IDL. They're not clickable when the IDL is the primary definition. Similarly, the definitions outside of IDL don't link back to the IDL. That's something that would need to change in Respec, rather than in this repository, and we'll need a precise description of where the links should go in every case in order to make that change.

g-ortuno:
Ah I see. This does makes things easier :smile:

PR #129, opened 2015-06-08
Expose all interfaces with user-visible instances, and make sure they start with "Bluetooth".

jyasskin:
This allows users to add helper functions to their prototypes without getting an
object first. This spec claims the window.Bluetooth* namespace.

I've included `Bluetooth`, the type of `navigator.bluetooth`, for symmetry, even though users can just modify `navigator.bluetooth` directly.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/expose-the-right-interfaces/index.html.


scheib:
LGTM

PR #134, opened 2015-06-25
Use HTTPS where possible

mathiasbynens:


index.html
```diff
@@ -3387,18 +3387,18 @@
         <dt>[[!ECMAScript]]</dt>
         <dd>
           <ul>
-            <li><a href="https://people.mozilla.org/~jorendorff/es6-draft.html#sec-array-objects"
+            <li><a href="http://ecma-international.org/ecma-262/6.0/#sec-array-objects"
```
jyasskin: This change doesn't match the title of your pull request, although it might be the right thing to do.

beaufortfrancois: I think that's why he added "where possible" in the Commit Message. We may want to merge this PR now. WDYT?

mathiasbynens: Yeah that’s why. HTTPS > HTTP, but canonical spec links > unofficial draft snapshots.

beaufortfrancois: @jyasskin (gentle ping)

jyasskin: Yeah, sorry for the delay. It's more that we should commit this as a separate commit titled "Switch links to the final ECMAScript 6 spec." I'll split it now.

jyasskin:
Merged in 62914209f9d201365c4434eea7e4ab10647ffb57 and 4693017e653a7e5fa059fc1433cab4d3cf7b8d29. Thanks!

Issue #133, opened 2015-06-21
Apply [EnforceRange] to the BluetoothUUID operations


jyasskin:
This would throw a TypeError for numeric values that aren't 32-bit UUID aliases.

This depends on https://www.w3.org/Bugs/Public/show_bug.cgi?id=28834, but we could willfully violate WebIDL and just make the UAs implement it.

jyasskin:
We should apply it to `canonicalUUID` first, since strings and objects convert to `0` there, but stay as non-UUID strings for the other functions.

jyasskin:
Also see https://www.w3.org/Bugs/Public/show_bug.cgi?id=28875

PR #136, opened 2015-07-01
Enforce that canonicalUUID() is called with an in-range alias.

jyasskin:
Issue #133 tracks applying this to the other functions taking unsigned long.
This also allows some non-integer arguments, due to WebIDL details.

https://codereview.chromium.org/1210173011/ shows the behavior.


scheib:
LGTM

PR #143, opened 2015-07-23
Throw NotFoundError when trying to look up a non-existent attribute.

### LINKED ISSUE #124: getPrimaryService (and similar) should throw if UUID is not found
After requestDevice succeeds for my set of services, an application believes it has access to a device that supports known services. Those services have known characteristics. Calling getPrimaryService(uuid1) and getCharacteristics(uuid2) should only have a promise fulfilled if services and characteristics existed and were obtained as expected. Applications shouldn't need to check for errors int he fulfillment of that promise:

```
navigator.bluetooth.requestDevicerequestDevice({
  filters: [ {services: [uuid1]} ]
}).then(function(device) {
  return device.connectGATT();
}).then(function(gattServer) {
  return gattServer.getPrimaryService(uuid1);
}).then(function(service) {
  return service.getCharacteristics(uuid2);
}).then(function(characteristics) {
  // Code here trusts that my requests have been fulfilled correctly.
  // If my uuids were correct, I know exactly the service, and that it
  // has certain number of characteristics with uuid2.
  // If for some reason they are not found, it is exceptional and
  // should have rejected a previous promise.
  characteristics[5]; // should exist and be valid without checking.
}).catch(... oh noes, normal exception handling ...);
```

- getCharacteristic requires a UUID, and it makes sense to reject if that uuid isn't found.
- getCharacteristics with a provided UUID should reject if not found for similar reasons.
- getCharacteristics without a provided UUID requires additional thought. I could go either way at the moment: I suppose if specific UUID isn't provided then the application doesn't have an expectation of what it will receive back, and an empty set may be valid. Consistency with "NotFound" == reject promise for the previous forms is reasonable as well.

jyasskin:
The common case seems to be navigating a tree with certain attributes expected
to be there, and null checks make that more complicated. It's still possible to
handle optional attributes by catching the exception.

We have less experience with the multi-attribute getters, but I'm doing them
consistently for now.

Fixes #124.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/throw-for-missing-attributes/index.html. @beaufortfrancois @scheib

### DIFF
--- a/index.html
+++ b/index.html
@@ -1616,8 +1616,8 @@ return navigator.bluetooth.requestDevice({
           </li>
           <li>
             Return the result of
-            <a>transforming</a> <var>promise</var> with a fulfilment handler that
-            returns <code>null</code> if its argument is empty
+            <a>transforming</a> <var>promise</var> with a fulfillment handler that
+            throws a <a>NotFoundError</a> if its argument is empty
             or returns the first (only) element of its argument.
           </li>
         </ol>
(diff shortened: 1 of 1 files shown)


scheib:
LGTM. Consistency wins here I think, and if expecting the possibility of no services being found there's little downside to having conditional logic in a then(,) vs then(if).

PR #144, opened 2015-07-23
Throw an exception if requestDevice() filters include no services

### LINKED ISSUE #142: requestDevice should fail loudly if it gets empty filters
Right now, [matches a filter](https://webbluetoothcg.github.io/web-bluetooth/#dfn-matches-a-filter) returns that no devices match if the filter's services are empty. (This encourages developers to only present users devices the site can actually use.) However, this turns out to be confusing to developers. Instead, `requestDevice()` should throw an exception for forbidden filter arguments.

jyasskin:
instead of silently returning no devices.

Fixes #142.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/loud-empty-filters/index.html. @beaufortfrancois @scheib

### DIFF
--- a/rationale.md
+++ b/rationale.md
@@ -32,6 +32,12 @@ page doesn't filter the devices that appear in the `requestDevice()` dialog,
 users could easily select a device the web page can't use, which is a bad
 experience.
 
+One reason for scanning without service UUID filters is to collect just
+advertising data from several nearby devices. The first version of Web Bluetooth
+is aimed at GATT communication with particular paired devices and doesn't
+support this use case. A subsequent version will most likely include such
+support, likely through a different entry point.
+
 ## Why so many `Get{,All}{Service,Characteristic,Descriptor}()` overloads?
 
 GATT provides two ways of finding primary services:
--- a/rationale.md
+++ b/rationale.md
@@ -24,6 +24,14 @@ This document attempts to explain why various decisions were made the way they w
 * While the <a>Server</a> role can work in a <a>Central</a> device,
   <a>Client</a> is more natural and is required for more use cases.
 
+## Why does `requestDevice()` require non-empty filters?
+
+In order to communicate with a device, that device needs to support a GATT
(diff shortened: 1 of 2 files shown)


rationale.md
```diff
@@ -24,6 +24,14 @@ This document attempts to explain why various decisions were made the way they w
 * While the <a>Server</a> role can work in a <a>Central</a> device,
   <a>Client</a> is more natural and is required for more use cases.
 
+## Why does `requestDevice()` require non-empty filters?
```
jyasskin: @kenchris commented on c189c7675459ec24a779ed7179912a2fc93e0cdf that "Maybe it should then be called `requestDeviceMatching(...)` or similar to make this more obvious?"

I'm not sure it needs to be more obvious than getting a TypeError when you forget to pass filters.

scheib: I think requestDevice is sufficient, and that adding Matching doesn't add value to warrant the name complexity.

rationale.md
```diff
@@ -24,6 +24,14 @@ This document attempts to explain why various decisions were made the way they w
...
+
+In order to communicate with a device, that device needs to support a GATT
+Service that the web page understands and has permission to access. If the web
+page doesn't filter the devices that appear in the `requestDevice()` dialog,
+users could easily select a device the web page can't use, which is a bad
+experience.
+
```
scheib: Possibly acknowledge that attempting to read any ad data from devices is a non-goal of the current spec, and that use cases for scanning and reading AD data deserve an API design specifically for those scenarios.

scheib:
LGTM, though I offer a possible further explanation https://github.com/WebBluetoothCG/web-bluetooth/pull/144/files#r35387445

rationale.md
beaufortfrancois: :+1:

rationale.md
jyasskin: How's this?

scheib: LGTM

PR #146, opened 2015-07-27
Add the GATT blacklist.

### LINKED ISSUE #24: Blacklisting various services, characteristics, and descriptors.
The reason for having the BluetoothGattCharacteristic.startNotifications|stopNotifications methods in the first place is to prevent apps from directly writing to the Client Characteristic Configuration (CCC below) descriptor (which is used to enable/disable notifications and indications from the characteristic). This raises the following two points:

- BluetoothGattDescriptor.writeValue should NOT allow writing to the CCC descriptor directly to prevent one application from disabling notifications/indications while another application is interested in them. Either that or the (CCC) descriptor perhaps shouldn't ever be exposed.
- It should be made clear to the implementation side that startNotifications and stopNotifications internally needs to keep a reference count of notification "sessions" of sort to manage the value of the CCC descriptor.

jyasskin:
Fixes #24.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/blacklists/index.html. The registry PR is at https://github.com/WebBluetoothCG/registries/pull/1. @scheib @marcoscaceres

### DIFF
--- a/index.html
+++ b/index.html
@@ -3048,15 +3048,24 @@ return navigator.bluetooth.requestDevice({
             </li>
             <li>
               If <var>line</var> consists of just a <a>valid UUID</a>,
-              add a mapping from that UUID to "<code>exclude</code>" in <var>result</var>.
+              let <var>uuid</var> be that UUID and
+              let <var>token</var> be "<code>exclude</code>".
             </li>
             <li>
               If <var>line</var> consists of a <a>valid UUID</a>, a space (U+0020),
               and one of the tokens "<code>exclude-reads</code>" or "<code>exclude-writes</code>",
-              add a mapping from that UUID, to the token.
+              let <var>uuid</var> be that UUID and
+              let <var>token</var> be that token.
             </li>
             <li>
               Otherwise, return an error and abort these steps.
+            </li>
+            <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2936,6 +3019,77 @@
...
+              continue to the next line.
+            </li>
+            <li>
+              If <var>line</var> consists of just a <a>valid UUID</a>,
+              add a mapping from that UUID to "<code>exclude</code>" in <var>result</var>.
+            </li>
+            <li>
+              If <var>line</var> consists of a <a>valid UUID</a>, a space (U+0020),
+              and one of the tokens "<code>exclude-reads</code>" or "<code>exclude-writes</code>",
+              add a mapping from that UUID, to the token.
```
scheib: Previous block mentions adding a mapping to result. This block likely should mention 'result' as well, or perhaps have 3 maps. To be very precise language should indicate what happens when a UUID appears multiple times with different tokens.

jyasskin: Done, still with a single map.

scheib:
LGTM

jyasskin:
Thanks. I'm going to wait to merge this until WebBluetoothCG/registries#1 is ready, so the spec never has a broken link.

Issue #32, opened 2014-08-21
`address` should be `instanceId`


jyasskin:
MacOS doesn't provide the BT address for a peripheral, just a generated [identifier](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBPeripheral_Class/#//apple_ref/occ/instp/CBPeripheral/identifier).

This may also help avoid confusion between the public address, static address, private addresses, and IRK in picking a device's "address".

PR #147, opened 2015-07-27
Rename BluetoothDevice.instanceID to BluetoothDevice.id.

### LINKED ISSUE #139: Fix "ID" capitalization
In https://github.com/w3c/web-nfc/issues/27#issuecomment-122013007 @domenic says "Id" should only have an initial capital. @annevk points out in https://github.com/jyasskin/web-bluetooth-1/commit/8154c44f9857336bb5f1d93a0aadfa3a1c19dc12#commitcomment-12208486 that I changed it the other way. https://github.com/w3ctag/design-principles/issues/15 doesn't have explicit guidance for this yet. We could avoid this by just giving `BluetoothDevice`s "id"s instead of "instanceId/D"s.

jyasskin:
This avoids questions about how to capitalize "ID". There aren't any other uses
of instanceID left, since #113 removed them, so we don't need this spelling to
be consistent.

Fixes #139.

Demo at https://rawgit.com/jyasskin/web-bluetooth-1/rename-instanceid/index.html.

### DIFF
--- a/index.html
+++ b/index.html
@@ -946,7 +946,7 @@ href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicV
           };
 
           interface BluetoothDevice {
-            readonly attribute DOMString instanceID;
+            readonly attribute DOMString id;
             readonly attribute DOMString? name;
             readonly attribute BluetoothAdvertisingData adData;
             readonly attribute unsigned long? deviceClass;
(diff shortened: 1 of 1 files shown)


jyasskin:
@jracle @armansito You both commented on #32 where we picked `instanceID` so FYI in case you want to object here.

scheib:
LGTM

PR #148, opened 2015-07-30
Unify the GATT tree navigation algorithms.

jyasskin:
Demo at https://rawgit.com/jyasskin/web-bluetooth-1/unify-tree-navigation/index.html.

The style is similar to https://streams.spec.whatwg.org/, but with named parameters and a little more formatting.

### DIFF
--- a/index.html
+++ b/index.html
@@ -96,6 +96,9 @@
           },
       };
     </script>
+    <style>
+      .argument-list { display: inline-block; vertical-align: top; }
+    </style>
   </head>
   <body>
     <section id='abstract'>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1421,112 +1424,164 @@
...
+              Let <var>promise</var> be the result.
+            </li>
+            <li>
+              Return the result of
+              <a>transforming</a> <var>promise</var> with a fulfillment handler that:
+              <ul>
+                <li>If its argument is empty, throws a <a>NotFoundError</a>,</li>
+                <li>
+                  Otherwise, if the <var>single</var> flag is set,
+                  returns the first (only) element of its argument.
```
g-ortuno: If you are checking for the flag here do you need to check when you're querying the cache as well?

index.html
```diff
@@ -1421,112 +1424,164 @@
...
+          </ol>
+        </section>
+
+        <section>
+          <h2>Navigating the Bluetooth Hierarchy</h2>
+
+          <p>
+            To <dfn>GetGATTChildren</dfn>(<span class="argument-list"><var>attribute</var>: GATT Attribute,<br>
+            <var>single</var>: boolean,<br>
+            <var>uuidCanonicalizer</var>: function,<br>
```
g-ortuno: Do we still need to pass the canonicalizer if we are passing the type already? We could get the canonicalizer from the type.

jyasskin: We could, but we'd need a switch on the type in this algorithm, and that takes as much space as just passing the canonicalizer in from the caller.

index.html
jyasskin: I do: 
> `<li>if the <var>single</var> flag is set, are the first of these.</li>`

g-ortuno:
LGTM

PR #149, opened 2015-07-30
Unify the BluetoothUUID.get* algorithms.

### LINKED ISSUE #131: Combine getService, getCharacteristic, getDescriptor algorithm
Could we combine those algorithms since they only differ in one line?

jyasskin:
Demo at https://rawgit.com/jyasskin/web-bluetooth-1/unify-uuid-algorithms/index.html#standardized-uuids

### DIFF
--- a/index.html
+++ b/index.html
@@ -2935,7 +2935,7 @@ return navigator.bluetooth.requestDevice({
             return <var>name</var> and abort these steps.
           </li>
           <li>
-            If <code><var>prefix</var>.<var>name</var></code>
+            If the string <code><var>prefix</var> + "." + <var>name</var></code>
             appears in <var>assigned numbers table</var>,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2918,7 +2935,8 @@
             return <var>name</var> and abort these steps.
           </li>
           <li>
-            If <code>org.bluetooth.service.<var>name</var></code> appears in [[!BLUETOOTH-ASSIGNED-SERVICES]],
```
scheib: I think this works, but... it's a touch odd to mix variables and the '.' literal this way. An alternative would be something like: If the string prefix + "." + name appears in...

scheib:
LGTM; Optional noted string concatenation comment.

index.html
jyasskin: Done.

PR #157, opened 2015-09-03
Start a specification for test-only APIs used in the Web Bluetooth tests.

jyasskin:
This matches the current Chromium test APIs (eventSender.keyDown and
testRunner.setBluetoothMockDataSet()) except:
* FailingConnectionsAdapter is omitted because the BT standard doesn't include
  any portable error codes.
* FailingGATTOperationsAdapter returns an ATT Error Response corresponding to
  its UUID, instead of the Chromium-specific error code used in the current
  Chromium tests.

@scheib @g-ortuno, can you double-check that I've matched the Chromium test APIs? This PR doesn't include the new interfaces I'm adding in https://codereview.chromium.org/1325953002/.

@yrliou @shuangMoz, we'd like your implementation to be able to share as many of our tests as possible. Please let us know what of this we need to change for it to work for you.

Rendered spec at https://rawgit.com/jyasskin/web-bluetooth-1/test-apis/tests/index.html.

### DIFF
--- a/tests/index.bs
+++ b/tests/index.bs
@@ -1,14 +1,13 @@
 <pre class='metadata'>
 Title: Testing Web Bluetooth
+Group: web-bluetooth-cg
 Status: CG-DRAFT
 ED: https://webbluetoothcg.github.io/web-bluetooth/tests
 Shortname: testing-web-bluetooth
 Level: 1
 Editor: See contributors on GitHub, , https://github.com/WebBluetoothCG/web-bluetooth/graphs/contributors
 Abstract: This document describes functions used in testing the Web Bluetooth API.
-
-Group: web-bluetooth-cg
-
+Status Text: <strong class="advisement">This specification does not yet reflect the consensus of the Web Bluetooth CG. It is presented by the Chromium developers in the hope that other implementers will find it useful. It may be withdrawn if they don't.</strong>
 Markup Shorthands: css no
 Link Defaults: html (dfn) trusted event/the body element
 Link Defaults: web-bluetooth (dfn) valid uuid
--- a/tests/index.bs
+++ b/tests/index.bs
@@ -34,6 +34,9 @@ spec: BLUETOOTH-ASSIGNED
     type: enum; urlPrefix: https://developer.bluetooth.org/gatt/
         urlPrefix: characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.
             text: org.bluetooth.characteristic.gap.device_name; url: gap.device_name.xml#
+        urlPrefix: descriptors/Pages/DescriptorViewer.aspx?u=org.bluetooth.descriptor.
+            text: org.bluetooth.descriptor.gatt.characteristic_user_description; url: gatt.characteristic_user_description.xml#
+            text: org.bluetooth.descriptor.gatt.characteristic_extended_properties; url: gatt.characteristic_extended_properties.xml#
         urlPrefix: services/Pages/ServiceViewer.aspx?u=org.bluetooth.service.
             text: org.bluetooth.service.battery_service; url: battery_service.xml#
             text: org.bluetooth.service.glucose; url: glucose.xml#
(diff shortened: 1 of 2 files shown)


tests/index.bs
```diff
@@ -0,0 +1,213 @@
...
+:: The UA fails to start a scan for devices.
+: <code>"GlucoseHeartRateAdapter"</code>
+:: The UA discovers a <a>HeartRateDevice</a> and then a <a>GlucoseDevice</a>.
+: <code>"MissingServiceGenericAccessAdapter"</code>
+:: The UA discovers a <a>MissingServiceGenericAccessDevice</a>.
+: <code>"MissingCharacteristicGenericAccessAdapter"</code>
+:: The UA discovers a <a>MissingCharacteristicGenericAccessDevice</a>.
+: <code>"GenericAccessAdapter"</code>
+:: The UA discovers a <a>GenericAccessDevice</a>.
+: <code>"FailingGATTOperationsAdapter"</code>
```
g-ortuno: What about FailingConnectionAdapter?

g-ortuno: Also should we have a test for failing to stop? I know we just ignore it, but we should make sure the code actually ignores it.

tests/index.bs
```diff
@@ -0,0 +1,213 @@
...
+    discovers a <a>BatteryDevice</a>.
+: <code>"EmptyAdapter"</code>
+:: No devices are nearby.
+: <code>"FailStartDiscoveryAdapter"</code>
+:: The UA fails to start a scan for devices.
+: <code>"GlucoseHeartRateAdapter"</code>
+:: The UA discovers a <a>HeartRateDevice</a> and then a <a>GlucoseDevice</a>.
+: <code>"MissingServiceGenericAccessAdapter"</code>
+:: The UA discovers a <a>MissingServiceGenericAccessDevice</a>.
+: <code>"MissingCharacteristicGenericAccessAdapter"</code>
```
g-ortuno: Will this document *only* list Data Sets that we've implemented? Otherwise we should include `MissingDescriptorGenericAccessAdapter` and have `GenericAccessDevice` inherit from `MissingDescriptorGenericAccessDevice`

tests/index.bs
```diff
@@ -0,0 +1,213 @@
...
+<h2 id="security-and-privacy">Security and privacy considerations</h2>
+
+These functions MUST NOT be exposed to web content. Only trusted testing
+code may access them.
+
+
+<h2 id="test-interfaces">Testing interfaces</h2>
+<pre class="idl">
+  partial interface Window {
+    readonly attribute TestRunner testRunner;
```
scheib: These are a bit WebKit history centric. We may wish to find entry points that are agreed upon by other browsers. I had an initial email chat with jgraham@ some months ago and he suggested we bring this up in public-test-infra@w3.org once we have a sturdy example to point to.

scheib: @jgraham

tests/index.bs
jyasskin: > What about FailingConnectionAdapter?

[FailingConnectionsAdapter](https://code.google.com/p/chromium/codesearch/#chromium/src/content/shell/browser/layout_test/layout_test_bluetooth_adapter_provider.h&l=154-172) holds devices that fail with each of the Chromium-specific [`ConnectErrorCode`s](https://code.google.com/p/chromium/codesearch/#chromium/src/device/bluetooth/bluetooth_device.h&l=88-96). We'll need to make those failure codes not-Chromium-specific before putting them here, and I don't see an obvious way to do that, unlike for the GATT errors. Any ideas?

> Also should we have a test for failing to stop? I know we just ignore it, but we should make sure the code actually ignores it.

We should, but I'd like to add extensions to Chromium's current API in a separate PR, so that we get a base as soon as possible.

tests/index.bs
jyasskin: I'm happy to add aspirational APIs later, but I think this PR should be just the sharable subset of what we've already done.

tests/index.bs
jyasskin: I'll post this PR on public-test-infra. I agree the names are weird, and I have no idea where they came from. In Chromium/Blink, we can probably polyfill any spec'ed interface names using `testharnessreport.js` (although that's also a weird location for polyfills).

tests/index.bs
g-ortuno: Keeping only implemented mocks sounds good.

tests/index.bs
g-ortuno: re: FailingConnections
Ahh good point.

re: Failing Stop
Sounds good.

tests/index.bs
scheib: Yes, testharnessreport.js could have been named testharnessbrowserspecific.js or some such -- perhaps it still can be. Alternatively the Bluetooth test suite could come with it's own bluetoothbrowsertestfixture.js that is to be provided by browsers and map these concepts.

jyasskin: We haven't gotten any response on public-test-infra or from Mozilla. That may mean there's no real interest in having cross-browser tests for this, or that Mozilla hasn't started implementing yet, and they won't be interested until they need some tests.

I'm inclined to check this in so it's there for anyone who wants to re-use our tests, even though it might never be used. I've added a warning in the document status that it's more speculative than many specifications. How's it look?

scheib: LGTM.

My thoughts: I believe that specifications should be complete under test,
and that if additional user-agent functionality is required to test a
specification it should be included in said specification. This is inspired
by the Bluetooth SIG's approach to controller design which specifies not
only the controller functionality but it's testing interfaces.

I don't think we need more consensus than we already have acquired for
other elements of the specification to add testing components to the
specification. General specification review hasn't had 100% coverage from
multiple vendors. So I believe the separation and the testing spec and the
tentative clauses are unnecessarily conservative.

PR #161, opened 2015-09-11
Chooser test apis

jyasskin:
These were implemented in Chrome's testing library in https://codereview.chromium.org/1325953002/. Demo at https://rawgit.com/jyasskin/web-bluetooth-1/chooser-test-apis/tests/index.html.

### DIFF
--- a/tests/index.bs
+++ b/tests/index.bs
@@ -93,14 +93,20 @@ code may access them.
 <pre class="idl">
   interface TestRunner {
     void setBluetoothMockDataSet(DOMString dataSetName);
+    void setBluetoothManualChooser();
+    sequence&lt;DOMString> getBluetoothManualChooserEvents();
+    void sendBluetoothManualChooserEvent(DOMString event, DOMString argument);
   };
 </pre>
 
 <h3 id="setBluetoothMockDataSet">setBluetoothMockDataSet</h3>
 
 When invoked, <dfn method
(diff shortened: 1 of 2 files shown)


tests/index.bs
```diff
@@ -208,6 +214,71 @@ read or written, the characteristic with UUID <code>errorUUID(which)</code>
...
+
+: The UA discovers that a device with {{BluetoothDevice/id}}
+    <var>device-id</var> is no longer nearby.
+:: <code>"remove-device(<var>device-id</var>)"</code>
+
+Note: {{TestRunner/getBluetoothManualChooserEvents()}} returns a
+<code>sequence&lt;DOMString></code> rather than the
+<code>{{Promise}}&lt;sequence&lt;DOMString>></code> that we'd normally prefer,
+for two reasons. First, the Chromium implementation has trouble returning a
+{{Promise}} from this interface, for bad reasons. Second, we aren't providing a
```
scheib: ;) The second reason is a bit weak (we could of course have send...Event return a promise if the first reason wasn't an issue). I think the first reason is valid enough.

scheib:
LGTM, though the one comment could be shortened as noted.

tests/index.bs
jyasskin: Fair enough. How's this?

scheib: LGTM

PR #167, opened 2015-09-23
Make getBluetoothManualChooserEvents asynchronous.

jyasskin:
This uses a callback rather than a promise because of an objection in
https://codereview.chromium.org/1351393002.

Preview at https://rawgit.com/jyasskin/web-bluetooth-1/async-get-events/tests/index.html#test-runner.


scheib:
LGTM

PR #174, opened 2015-10-14
Allow filtering by device name and name prefix in requestDevice().

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/new-scan-filters/index.html#device-discovery.

I think I got all the places that assumed filters only had services, but please let me know if I missed any.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -103,6 +103,7 @@ spec: dom-ls; urlPrefix: http://dom.spec.whatwg.org/#; type: dfn
 spec: ECMAScript; urlPrefix: http://ecma-international.org/ecma-262/6.0/#
     type: method
         text: Array.prototype.map; url: sec-array.prototype.map
+        text: String.prototype.startsWith; url: sec-string.prototype.startswith
     type: interface
         text: Array; url: sec-array-objects
         text: ArrayBuffer; url: sec-arraybuffer-constructor
@@ -112,6 +113,10 @@ spec: ECMAScript; urlPrefix: http://ecma-international.org/ecma-262/6.0/#
         text: Set; url: sec-set-objects
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
         text: TypedArray; url: sec-typedarray-constructors
+
+spec: encoding; urlPrefix: https://encoding.spec.whatwg.org/#
+    type: dfn
+        text: utf-8 decode without BOM; url: utf-8-decode-without-bom
 
 spec: powerful-features; urlPrefix: https://w3c.github.io/webappsec/specs/powerfulfeatures/#
     type: dfn
@@ -527,6 +532,8 @@ spec: html
   <pre class="idl">
     dictionary BluetoothScanFilter {
       sequence&lt;BluetoothServiceUUID> services;
+      DOMString name;
+      DOMString namePrefix;
     };
 
     dictionary RequestDeviceOptions {
@@ -545,11 +552,24 @@ spec: html
     <p>
       {{Bluetooth/requestDevice(options)}} asks the user
       to grant this origin access to a device
-      that <a>matches a filter</a> in <code>options.<dfn dict-member for="RequestDeviceOptions">filters</dfn></code>.
-      The user will be shown devices that support
-      <em>all</em> the GATT service UUIDs in the <dfn dict-member for="BluetoothScanFilter">services</dfn> list
-      of <em>any</em> {{BluetoothScanFilter}} in <code>filters</code>.
+      that <a>matches any filter</a> in <code>options.<dfn dict-member for="RequestDeviceOptions">filters</dfn></code>.
+      To <a>match a filter</a>, the device has to:
     </p>
+    <ul>
+      <li>
+        support <em>all</em> the GATT service UUIDs
+        in the <dfn dict-member for="BluetoothScanFilter">services</dfn> list
+        if that member is present,
+      </li>
+      <li>
+        have a name equal to <dfn dict-member for="BluetoothScanFilter">name</dfn>
+        if that member is present, and
+      </li>
+      <li>
+        have a name starting with <dfn dict-member for="BluetoothScanFilter">namePrefix</dfn>
+        if that member is present.
+      </li>
+    </ul>
     <p>
       After the user selects a device to pair with this origin,
       the origin is allowed to access to any service whose UUID was listed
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
I was surprised to see that we allow `namePrefix` to be an empty string. 
This means that _If the web page doesn't filter the devices that appear in the requestDevice() dialog, users could easily select a device the web page can't use, which is a bad experience._ in https://github.com/WebBluetoothCG/web-bluetooth/blob/gh-pages/rationale.md#why-does-requestdevice-require-non-empty-filters is not true anymore.

jyasskin:
Good point. The new patch requires a non-empty `namePrefix` and that some member of each filter is present.

index.bs
```diff
@@ -624,14 +644,129 @@ spec: html
...
+            [{namePrefix: ""}]
+          </pre>
+        </td>
+        <td><i>TypeError</i></td>
+        <td>A filter must restrict the devices in some way.</td>
+      </tr>
+    </table>
+
+    <p>
+      Remember to use {{RequestDeviceOptions/optionalServices}} to get access to services
```
g-ortuno: This seems like it will be a common pitfall. Can we make it stand out more?

index.bs
```diff
@@ -898,6 +1067,8 @@ spec: html
         An optional partial or complete <a>Bluetooth Device Name</a>.
         A device has a partial name when the <a>Shortened Local Name</a> AD data was received,
         but the full name hasn't been read yet.
+        When the <a>Bluetooth Device Name</a> is encoded as UTF-8 and
+        converted to a DOMString using the <a>utf-8 decode without BOM</a> algorithm.
```
g-ortuno: I think this is a sentence fragment?

g-ortuno:
Two comments. Otherwise LGTM

scheib:
I like that a solution for names is being worked on. But:
- Why support a substring match only for prefix? If a substring match is supported why not for any portion of the string? I agree that most devices seem to use a prefix, but it's hard to be exhaustive. "Red Fish", "Blue Fish" seem reasonable names too. 
- The substring can be as short as a single character. This starts to have fairly diminishing returns as a filter for security purposes. It makes me wonder why not support an empty filter too.

index.bs
```diff
@@ -624,14 +644,129 @@ spec: html
...
+        <td>A filter must restrict the devices in some way.</td>
+      </tr>
+      <tr>
+        <td>
+          <pre highlight="js">
+            [{namePrefix: ""}]
+          </pre>
+        </td>
+        <td><i>TypeError</i></td>
+        <td>A filter must restrict the devices in some way.</td>
```
scheib: 'namePrefix', if present, must be non-empty to filter devices.

index.bs
```diff
@@ -545,11 +551,24 @@ spec: html
...
+      </li>
+      <li>
+        have a name equal to <dfn dict-member for="BluetoothScanFilter">name</dfn>
+        if that member is present, and
+      </li>
+      <li>
+        have a name starting with <dfn dict-member for="BluetoothScanFilter">namePrefix</dfn>
+        if that member is present.
+      </li>
+    </ul>
```
g-ortuno: Should we enforce a max length on the name and namePrefix? BLUETOOTH SPECIFICATION Version 4.2 [Vol 3, Part C] 3.2.2.3 Representation:
`The Bluetooth device name can be up to 248 bytes`

index.bs
jyasskin: I've moved it to the "Bluetooth members" note. How's that?

index.bs
jyasskin: Yep. I think I meant to write it without the "When".

index.bs

index.bs
jyasskin: Name filters longer than that will naturally not match anything. We could warn, but I'm not sure it makes sense to throw an error?

jyasskin:
> Why support a substring match only for prefix? If a substring match is supported why not for any portion of the string? I agree that most devices seem to use a prefix, but it's hard to be exhaustive. "Red Fish", "Blue Fish" seem reasonable names too.

The best thing to do might be to support a regex, but IIRC we'd have
trouble running that regex in the browser process where we need to
filter things.

Advertisements send a prefix of the device name, so we can be more
sure a device doesn't match a prefix than a substring. If we start
connecting to devices to check their filters, this would let us
connect to fewer, but we don't have plans to connect, so maybe it
doesn't matter.

> The substring can be as short as a single character. This starts to have fairly diminishing returns as a filter for security purposes. It makes me wonder why not support an empty filter too.

Yeah; maybe we should back off of insisting that folks filter? If they
want to produce a bad experience by letting users pick devices they
can't use, let them?

scheib:
>
>
> Advertisements send a prefix of the device name, so we can be more
> sure a device doesn't match a prefix than a substring.


That's pretty compelling. Connecting to devices to improve filtering goes
against our privacy model in the UI.


> Yeah; maybe we should back off of insisting that folks filter? If they
> want to produce a bad experience by letting users pick devices they
> can't use, let them?


I'm close to saying we should go for this idea. If we can think of how, it
would be good to make it API obvious that this isn't desirable.. meaning
you should have to write something more or as much of a filter in order to
do a wildcard filter.
requestDevice({ filters: [{
ok_I_know_I_should_have_one_but_for_my_device_use_case_there_just_isn't_one:
true }]

index.bs
g-ortuno: palmer@chromium.org, who is doing the security review for the CL, asked if there are any constraints we should enforce [Link](https://codereview.chromium.org/1415533006/diff/80001/content/browser/bluetooth/bluetooth_dispatcher_host.cc). Device Name length seems like something we can enforce.

jyasskin: Done, although note that Chrome can avoid sending the IPC for shorter names than the spec's limit, since we know we won't look past the advertising packet.

jyasskin:
As long as the browser's device implements the Privacy Feature (which Chrome, Android, and Mac do, at least), I don't think connecting hurts privacy. It mainly hurts performance.

If we want to back off of insisting that filters screen out some devices, I think the main benefit would be simplifying the specification. Trying to make wildcard filters verbose blocks all the simplifications we'd otherwise be able to make, so I think we should go all the way or not do it at all.

jyasskin:
I'm just going to commit this for now. We can change it later if we think there's a better API.

PR #177, opened 2015-10-19
Fixed nit in implementation status

beaufortfrancois:
R=@scheib


scheib:
Doh, my bad, but the fix is to change it to 5.

PR #187, opened 2015-11-09
Block an empty list of filters.

jyasskin:
@g-ortuno https://rawgit.com/jyasskin/web-bluetooth-1/reject-empty-filters/index.html#dom-bluetooth-requestdevice

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -798,6 +798,11 @@ spec: html
       do the following substeps:
       <ol>
         <li>
+          If <code><var>options</var>.filters.length === 0</code>,
+          <a>reject</a> <var>promise</var> with a {{TypeError}}
+          and abort these steps.
+        </li>
+        <li>
           Let <var>uuidFilters</var> be a new {{Array}} and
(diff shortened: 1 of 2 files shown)


g-ortuno:
Should we add an example of this to [Example 3](https://webbluetoothcg.github.io/web-bluetooth/#example-filter-by-name)?

jyasskin:
Sure. Do you think I should also disallow filters that specify both a name and a namePrefix? Since that's either redundant or eliminates all devices.

g-ortuno:
This LGTM. Re: Not allowing both `name` and `namePrefix`. I'm not opposed. But maybe we should reconsider our approach so we only expose one field. Talking with @scheib he mentioned that an ignore case option would be useful so maybe expose prefix as an options as well. Something like: 
```js
  filters: [{
    name: {
      str: 'somename',
      ignoreCase: true,
      prefix: true,
      postfix: ,
      fragment: ,
      ...
    }
  }]
});
```
It would allow developers more options for searching by name and keep the implementation simple.

Issue #189, opened 2015-12-04
Remove BluetoothDevice.paired


jyasskin:
It looks like CoreBluetooth has no way to get this information, so we can't expose it universally. It's a holdover from `chrome.bluetooth`.

beaufortfrancois:
Work in progress at https://codereview.chromium.org/1564953003/

beaufortfrancois:
https://codereview.chromium.org/1564953003/ has landed.

PR #197, opened 2016-01-12
Remove BluetoothDevice.paired

beaufortfrancois:
R=@jyasskin

Fix https://github.com/WebBluetoothCG/web-bluetooth/issues/189


jyasskin:
LGTM. We can also remove the 'bonded flag' definition, but I'll do that in a followup.

PR #198, opened 2016-01-14
Remove all mentions of "script execution environment", which no longer exists in HTML.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/remove-script-execution-environment/index.html

@bzbarsky It looks like other specs leave platform objects' global environments implied rather than explicitly wired, so I'm following that convention here. I wound up using internal slots on the `Bluetooth` instance in nearly all cases, so I only needed to use that implied environment in one place: in https://rawgit.com/jyasskin/web-bluetooth-1/remove-script-execution-environment/index.html#service-change-events to find an event loop.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -100,7 +100,9 @@ spec: BLUETOOTH-ASSIGNED
 spec: dom-ls; urlPrefix: http://dom.spec.whatwg.org/#; type: dfn
     text: participate in a tree; url: concept-tree-participate
 
-spec: ECMAScript; urlPrefix: http://ecma-international.org/ecma-262/6.0/#
+spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
+    type: dfn
+        text: Realm; url: sec-code-realms
     type: method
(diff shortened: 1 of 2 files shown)


bzbarsky:
> It looks like other specs leave platform objects' global environments implied rather than explicitly wired

Other specs typically get it wrong.  Though there _is_ talk about adding some sort of defaults to IDL where an object gotten off another object will just be associated with the same global unless something says otherwise, mostly to work around everyone not even thinking about the problem.  In practice what this means is that right now browsers often disagree once you start trying to test what they're actually doing.

jyasskin:
@annevk seems to think that WebIDL will wind up adding those defaults. I don't really want to be the first to adopt a temporary state, so I'd rather let this settle out, and then add anything necessary to this spec if we're not totally covered by the defaults.

annevk:
Yeah, this is https://www.w3.org/Bugs/Public/show_bug.cgi?id=24652 as far as I can tell.

PR #200, opened 2016-01-16
Make readValue's .then() and event handlers execute in the same order they're mentioned in the spec.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/read-value-ordering/index.html

@g-ortuno I'm being clever here with the adjacent tasks. Could you double-check that it looks reasonable to implement? `BluetoothDispatcher::OnReadValueSuccess` currently just resolves the promise, and doesn't fire any events. To avoid letting another value update come in before the event dispatches, we'd either need to run a microtask checkpoint (which I could just put in the spec...), queue a pair of tasks from `OnReadValueSuccess`, or send two adjacent IPCs to get the two tasks queued.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2309,12 +2309,21 @@ spec: html
           <li>
             <a>Resolve</a> <var>promise</var> with <code>this.value</code>.
           </li>
-          <li>
-            <a>Fire an event</a> named {{characteristicvaluechanged}}
-            with its <code>bubbles</code> attribute initialized to <code>true</code>
-            at <code>this</code>.
-          </li>
         </ol>
+      </li>
+      <li>
+        <p>
+          <a>Queue a task</a> to <a>fire an event</a> named {{characteristicvaluechanged}}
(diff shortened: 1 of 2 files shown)


g-ortuno:
We should be good on the implementation side.

More details: We send the events through [GattCharacteristicValueChanged](https://code.google.com/p/chromium/codesearch#chromium/src/content/browser/bluetooth/bluetooth_dispatcher_host.cc&l=563) where we post a task to notify the dispatcher that a characteristic changed; this way we resolve the promise first. In order to avoid sending unnecessary characteristic changed events to the dispatcher, the dispatcher host only sends events after a listener has been added to the characteristic, which requires another message from the dispatcher to the host. It is in between this message that the js context could miss an event. One solution we considered is just sending all messages to the dispatcher that way we avoid that extra message and no events get lost.

g-ortuno:
lgtm

PR #201, opened 2016-01-19
Fix #172: Return DataViews instead of ArrayBuffers.

### LINKED ISSUE #172: Returns DataView instead of ArrayBuffer on readValue
I've realized that all the `arrayBuffer` I receive from `readValue` are always wrapped around a `DataView` and I was wondering if we could simply return a `DataView` instead of an `ArrayBuffer` then.

Instead of 

```js
characteristic.readValue().then(buffer => {
  let data = new DataView(buffer);
  let foo = data.getUint8(0);
  ...
```

We would have something simpler like

```js
characteristic.readValue().then(data => {
  let foo = data.getUint8(0);
  // We can always access ArrayBuffer if we need to with data.buffer
  ...

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/data-view/index.html.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -256,8 +256,8 @@ spec: html
             return Promise.resolve();
           }
           return characteristic.<a for="BluetoothGATTCharacteristic">readValue()</a>
-          .then(sensorLocationArray => {
-            let sensorLocation = sensorLocationArray.getUint8(0);
+          .then(sensorLocationData => {
+            let sensorLocation = sensorLocationData.getUint8(0);
             switch (sensorLocation) {
               case 0: return 'Other';
               case 1: return 'Chest';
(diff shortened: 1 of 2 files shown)


g-ortuno:
You need to change the examples to use the DataView directly.

jyasskin:
I fixed one last use of `new Uint8Array`, but I think I got everything else. What are you seeing?

g-ortuno:
`parseHeartRate` "`parseHeartRate() would be defined using the heart_rate_measurement documentation to read the ArrayBuffer stored in a BluetoothGATTCharacteristic's value field`" and the code to implement it.

jyasskin:
I got that one. Search https://rawgit.com/jyasskin/web-bluetooth-1/data-view/index.html.

g-ortuno:
Ah my bad. LGTM.

beaufortfrancois:
`s/sensorLocationArray/sensorLocation/g` as it's not an array buffer anymore.

jyasskin:
I went with `sensorLocationData` so it doesn't collide with the variable that was initialized with `let sensorLocation = sensorLocationArray.getUint8(0);`.

PR #202, opened 2016-01-23
Fix #182: Filter services with the allowed services list.

### LINKED ISSUE #182: Take the allowed services list into account when returning services

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/filter-services/index.html#dom-bluetoothgattremoteserver-getprimaryservice.

This returns `NotFoundError` for out-of-bounds requests, rather than a more distinctive `SecurityError`. Do folks have a preference on that?

It also only enforces the allowed-services-list on primary services: once a primary service is allowed, sites have access to everything inside it, including other services that it includes.


g-ortuno:
To me it seems like `SecurityError` leaks the services to the page, but seems more useful to developers that forgot to add their services to `optionalServices`.

jyasskin:
As long as the `SecurityError` happens for *any* disallowed service, and not just ones that are actually present, I think we don't leak any information. I'll make that change.

g-ortuno:
lgtm

PR #203, opened 2016-01-25
Merge newly-granted allowed services into the stored list.

jyasskin:
This builds on #202.

Preview at https://rawgit.com/jyasskin/web-bluetooth-1/update-services/index.html.


index.bs
```diff
@@ -1297,7 +1306,7 @@ spec: html
         readonly attribute unsigned long? productID;
         readonly attribute unsigned long? productVersion;
         readonly attribute BluetoothGATTRemoteServer? gattServer;
-        readonly attribute UUID[] uuids;
+        readonly attribute FrozenArray&lt;UUID> uuids;
```
g-ortuno: According to the spec `[BluetoothDevice.]uuids lists the UUIDs of GATT services known to be on the device, that the current origin is allowed to access.` This makes me thing that this list needs to be modified whenever a new service is added to the device. If so how can this be a FrozenArray? On the other hand the spec never updates this list so maybe the definition of uuids is wrong?

jyasskin: https://rawgit.com/jyasskin/web-bluetooth-1/update-services/index.html#dom-bluetoothdevice-uuids says that when `uuids` is gotten, the UA has to return a FrozenArray that's the intersection of two internal slots. When the user allows more services, that updates the list that [`BluetoothDevice@[[allowedServices]]`](https://rawgit.com/jyasskin/web-bluetooth-1/update-services/index.html#dom-bluetoothdevice-allowedservices-slot) refers to, which leads to `uuids` returning a different value.

Right now, my wording doesn't say whether it's a new or cached FrozenArray that's returned. Returning a new one would mean that `device.uuids !== device.uuids`, which folks tend not to like. Caching fixes that, but specifying the cache makes it more likely that I'll miss an update if a new part of the spec starts modifying `[[allowedServices]]`. Maybe I should just specify the cache and be careful to update everything?

jyasskin: Ugh, threading issues mean I need to post a task to update each `BluetoothDevice` anyway. Update coming...

jyasskin: Ok, changes to the allowed services now post a task to update the `BluetoothDevice`s and explicitly handle the `uuids` cache.

index.bs
```diff
@@ -1641,42 +1757,92 @@ spec: html
...
+            <td><dfn>\[[unfilteredData]]</dfn></td>
+            <td><code>new {{Map}}()</code></td>
+            <td>
+              The advertised service data, converted to {{DataView}}s.
+            </td>
+          </tr>
+        </table>
+
+        <p>
+          The <a>map entries</a> of a {{BluetoothServiceDataMap}} instance are
```
g-ortuno: Should this have similar steps to `BluetoothDevice.uuids`?

index.bs
```diff
@@ -2954,6 +3124,8 @@ spec: html
             </li>
             <li>
               For each <a>Service</a> in <var>addedEntities</var>,
+              if the Service's UUID is in
```
g-ortuno: Maybe I'm missing something outside of this patch but if I read this correctly if a new service is added, `BluetoothDevice.uuids` would still return the old services, because cachedAllowedServices is still equal to allowedServices.

index.bs
jyasskin: I don't think so: [`maplike` types](https://heycam.github.io/webidl/#idl-maplike) can change their values after they're constructed without changing their identities, while arrays are a fundamental Javascript type for which the only way to prevent users from changing them is to freeze them from UA changes too. 

@bzbarsky (editor of WebIDL), is [this](https://rawgit.com/jyasskin/web-bluetooth-1/update-services/index.html#bluetoothservicedatamap) how `maplike` is supposed to work? The only other use I can find (http://webaudio.github.io/web-midi-api/) doesn't specify its map entries at all.

bzbarsky: Yes, the UA can change the contents of a `maplike`, and the page can only change them if it's not a `readonly maplike`.  You have to specify when they're allowed to change, obviously, but apart from that no obvious constraints on it.

index.bs
jyasskin: Argh, yes. Fixed. Updating the unfiltered UUIDs was also broken, so I fixed that in the [service-changed section](https://rawgit.com/jyasskin/web-bluetooth-1/update-services/index.html#service-change-events).

g-ortuno:
lgtm

Issue #114, opened 2015-05-20
Describe what happens in response to disconnection (a lost ATT Bearer)


jyasskin:
https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-gattserver describes `BluetoothDevice.gattServer` becoming `null`, and [`BluetoothGATTRemoteServer.connected`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothgattremoteserver-connected) needs to become false.

I'm not certain that `BluetoothDevice.gattServer` should actually become null after the first connection, and if it does then maybe it should go back to the original object on reconnection?

beaufortfrancois:
I agree with you it shouldn't become `null`. The `connected` set to `false` should be enough especially if we add connect/disconnect Bluetooth events (https://github.com/WebBluetoothCG/web-bluetooth/issues/194)

PR #206, opened 2016-01-28
Add a gattserverdisconnected event to report when a page needs to reconnect.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/disconnect-event/index.html

I'm using `gattserverdisconnected` to distinguish this from non-gatt connections and from clients that disconnect once we implement peripheral mode.

This doesn't fully address #114.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -552,6 +552,7 @@ spec: webidl
       Promise&lt;BluetoothDevice> requestDevice(RequestDeviceOptions options);
     };
     Bluetooth implements EventTarget;
+    Bluetooth implements BluetoothDeviceEventHandlers;
     Bluetooth implements CharacteristicEventHandlers;
     Bluetooth implements ServiceEventHandlers;
   </pre>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -3186,6 +3228,18 @@ spec: webidl
 
       <pre class="idl">
         [NoInterfaceObject]
+        interface BluetoothDeviceEventHandlers {
```
g-ortuno: Would it make more sense to define BluetoothGATTServerEventHandlers? The interface would include all the events that happen at the GATT Server level i.e. `serviceadded`, `serviceremoved`, `gattserverdisconnect`.

jyasskin: The `ServiceEventHandlers` actually fire on the affected Service and then bubble to the `BluetoothDevice`, while `gattserverdisconnected` doesn't have a Service to fire on and has to target the `BluetoothDevice` directly.

I thought about calling this `BluetoothGATTRemoteServerEventHandlers`, but then thought that might be misleading since the events don't actually target the `BluetoothGATTRemoteServer` object. 

Targeting the server object would require that we make it participate in the tree, and it doesn't actually give us much in terms of specificity since devices and gatt servers are 1-1.

g-ortuno: Makes sense. Thanks!

g-ortuno:
lgtm

g-ortuno:
qq: what would be the name of the event for a local gatt server disconnection?

jyasskin:
I think that'll be `gattclientdisconnected`.

g-ortuno:
https://crbug.com/582537

PR #208, opened 2016-01-29
Suggest UAs disconnect when a BluetoothDevice is gc'ed.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/disconnect-on-gc/index.html.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -157,6 +157,7 @@ spec: dom-ls
 spec: html
     type: dfn
         text: allowed to show a popup
+        text: browsing context
         text: environment settings object
         text: event handler idl attribute
         text: global object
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2129,6 +2133,20 @@ spec: webidl
...
     <p>
+      When no ECMAScript code can
+      observe an instance of {{BluetoothRemoteGATTServer}} <var>server</var> anymore,
+      the UA SHOULD run <code><var>server</var>.{{BluetoothRemoteGATTServer/disconnect()}}</code>.
+      <span class="note">
+        Because {{BluetoothDevice}} instances are stored in
+        <code>navigator.bluetooth.{{Bluetooth/[[deviceInstanceMap]]}}</code>,
+        this can't happen at least until navigation releases the global object or
+        closing the tab or window destroys the <a>browsing context</a>.
+        Disconnecting on garbage collection ensures that
```
g-ortuno: These two sentences seem orthogonal. I think they should be two separate notes.

jyasskin: Sure. All in 1 paragraph or in 3 different paragraphs?

g-ortuno: One works for me. I just saw both sentences be in the same note and kept trying to find the relation between them.

g-ortuno:
lgtm

index.bs

Issue #175, opened 2015-10-15
readValue()'s event and Promise handlers should execute in an order matching the spec


jyasskin:
[readValue()](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothgattcharacteristic-readvalue) queues a task to:

1. Resolve a promise and
1. Fire an event.

The promise's `.then()` handlers run on the microtask loop, but [event dispatching](https://dom.spec.whatwg.org/#concept-event-dispatch) happens synchronously, leading to an order of operations that's not clear from the spec. We should either reverse the order in the spec or queue another task for the event to make sure it happens after the promise.

I'm inclined to queue another task, although that does open the possibility that multiple readValue() calls will complete, and the intermediate values won't be reflected during any event handlers.

Issue #171, opened 2015-10-05
Should GetGATTChildren required connected == true?


scheib:
getPrimaryService and other methods rely on GetGATTChildren algorithm to obtain GATT objects. Should they fail if not connected?

The Chromium implementation currently attempts to obtain this information even when not connected. It could be cached, so it is possible to allow implementations to attempt to service these requests, but it would be cleaner to specify either that they must attempt cached results or reject with a known error.

jyasskin:
Probably yes. This is also related to #114. I think we have several holes around handling disconnected devices.

jyasskin:
I think I'd rather fail fast on a disconnected device, probably with a `NetworkError`, rather than allowing cached results.

g-ortuno:
This seems like something that should be added to the [Populate Cache Algorithm](https://webbluetoothcg.github.io/web-bluetooth/#populate-the-bluetooth-cache). Something like: If device.connected == false then return a Network Error.

PR #209, opened 2016-02-02
Enforce that the current realm is connected while running GATT procedures.

### LINKED ISSUE #114: Describe what happens in response to disconnection (a lost ATT Bearer)

jyasskin:
This change makes it so that a disconnect() call takes effect immediately and
prevents any subsequent changes to member variables from a returning GATT
procedure, even if the GATT procedure started before the disconnection. It also
forces all active requests to fail with a NetworkError so that no remnants of
the connection keep running after it's terminated.

This reverses the readValue() promise and event order from #175, in order to
avoid needing to perform a microtask checkpoint inside the algorithm.

Fixes #114 and #171.

Preview at https://rawgit.com/jyasskin/web-bluetooth-1/enforce-connectedness/index.html.


jyasskin:
@yrliou @BruceSunMozilla this PR affects how you'll implement disconnection.

index.bs
```diff
@@ -2196,6 +2244,65 @@ spec: webidl
     </ol>
 
     <p>
+      <span class="note">
+        Algorithms need to fail if
+        their {{BluetoothRemoteGATTServer}} was disconnected while they were running,
+        even if the UA stays connected the whole time
+        and the {{BluetoothRemoteGATTServer}} is subsequently re-connected before they finish.
+        We wrap the returned {{Promise}} to accomplish this.
+      </span>
```
g-ortuno: Should we put the note in a separate paragraph?

index.bs
```diff
@@ -2196,6 +2244,65 @@ spec: webidl
...
+      To create a <var>gattServer</var>-<dfn>connection-checking wrapper</dfn>
```
g-ortuno: Do you mean `gatt`? We don't have `gattServer` anymore.

index.bs
```diff
@@ -2196,6 +2244,65 @@ spec: webidl
...
+      around a {{Promise}} <var>promise</var>, the UA MUST:
```
g-ortuno: I wonder if we really need this wrapper. At first it makes sense, this way we don't have to check `connected` in each function. But now, looking a the functions, it makes less sense. We basically check `connected` in all but one of them. I think explicitly checking `connected` makes more sense that the wrapper.

index.bs
jyasskin: We removed `BluetoothDevice.gattServer`, but `BluetoothDevice.gatt` still refers to a gatt server, so I like using the longer name for the variable.

index.bs
jyasskin: The wrapper manages adding and removing `promise` from the [`[[activeAlgorithms]]`](https://rawgit.com/jyasskin/web-bluetooth-1/enforce-connectedness/index.html#dom-bluetoothremotegattserver-activealgorithms-slot) slot, in addition to effectively checking `connected`. 
Repeating the "add to `[[activeAlgorithms]]`" and "if not in `[[activeAlgorithms]]` fail with `NetworkError`" text isn't so bad, but I wouldn't want to repeat the removal code in each algorithm.
However, I could just leave promises in `[[activeAlgorithms]]` in the spec, and rely on the implementation to optimize away the memory leak. What do you think?

g-ortuno: Do we need [[activeAlgorithms]] if we manually check `connected` in each function?

jyasskin: Yes; its purpose is to catch pages that disconnect and reconnect while an algorithm is in progress. I tried to say this in the note just before https://rawgit.com/jyasskin/web-bluetooth-1/enforce-connectedness/index.html#connection-checking-wrapper; do you see a way to make it clearer?

index.bs

g-ortuno:
lgtm

Emill:
As discussed in https://github.com/WebBluetoothCG/web-bluetooth/issues/274#issuecomment-268266639, I don't like the order "fire event _then_ resolve promise" since that makes it impossible to distinguish notifications from read responses.

PR #210, opened 2016-02-02
Define how to revoke access to a device within an origin.

### LINKED ISSUE #116: Describe what happens when the UA revokes access to a device.
http://webbluetoothcg.github.io/web-bluetooth/#dfn-allowed-devices-map allows UAs to revoke access at any time, but doesn't yet specify what happens as a result. We could do this by adding several checks against the `allowed devices map` in each algorithm, but there might be a more elegant option.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/revoke-access/index.html. To work correctly, this depends on #209.

Revoking access disconnects the device and makes sure future connections to
that device's objects fail. As long as other algorithms check for a connection
before succeeding, they'll always fail on a revoked object.

Fixes #116.


index.bs
```diff
@@ -1232,20 +1232,14 @@ spec: webidl
...
-      The UA MAY remove devices from the <a>allowed devices map</a> at any time
+      The UA MAY <a>revoke Bluetooth access</a> to a device for an origin at any time
       based on signals from the user.
-      <span class="issue">
-        This needs a definition involving
-        removing {{BluetoothDevice}} instances from {{Bluetooth/[[deviceInstanceMap]]}}s
-        and clearing out their {{[[representedDevice]]}} fields.
+      <span class="note">
+        For example, if the user chooses not to remember access,
+        the UA might remove a device when the tab that was granted access to it is closed.
```
g-ortuno: Since you are removing the previous instance of `remove devices from allowed devices map` it becomes less clear what you mean when you say `remove a device`.

jyasskin: Good point; done.

g-ortuno:
As it is right now, if we revoke a permission in the middle of an operation you would just get a network error because of disconnection. Should we maybe reject with a security error? Not sure we need it in the spec but we will probably need it in the implementation.

jyasskin:
I think it's better to just get a `NetworkError`. I want unusual things like revocation to look like something the page is handling anyway, like out-of-range, so that pages don't crash when the unusual thing happens.

g-ortuno:
Ah ok. So we would still reject with an error on the implementation but we would keep that out of the spec?

jyasskin:
I think we'd have the implementation reject with a `NetworkError` too. It _might_ have a distinct message, but messages aren't in the spec.

g-ortuno:
OK. LGTM

Issue #215, opened 2016-02-24
Consider an event when a device is physically disconnected


jyasskin:
Calling [`BluetoothRemoteGATTServer.disconnect()`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothremotegattserver-disconnect) disconnects the current Realm and allows the UA to physically disconnect if no other Realm is connected. [`gattserverdisconnected`](https://webbluetoothcg.github.io/web-bluetooth/#eventdef-bluetoothdevice-gattserverdisconnected) in turn reports if the current Realm was disconnected without asking to be, but this doesn't say whether the device was [revoked](https://github.com/WebBluetoothCG/web-bluetooth/pull/210) or the physical connection was lost (and I think we want to preserve that ambiguity).

@beaufortfrancois reports that some devices treat physical disconnection as a semantic actions, and change their state in response, and web pages want to convey that to the user. Is that beyond the BT4.0 behavior that peripherals can have only one physical connection?

We could add a `.physicallyConnected` attribute to the `gattserverdisconnected` event, and use that to distinguish Realm vs UA disconnections (treating the UA as including its OS/platform here). I don't think we want to let a Realm force the UA to disconnect, since that can interfere with other Realms, but the information leak from telling a Realm whether other Realms are connected seems small enough to accept.

Thoughts?

scheib:
The use case I heard, a site wanting to know when a peripheral disconnects,
seems to be satisfied with a gattserverdisconnected message without more
information.

If a disconnection happens at any general time, the site can explain the
connection was lost, and prompt to reconnect. However, when the site
performs an operation that it expects will result in a disconnection upon
completion, it can wait for gattserverdisconnected and then show a
confirmation.

The chance that a false disconnection and confirmation will occur due to
other reasons seems low.

jyasskin:
@beaufortfrancois Does the GATT operation cause the remote device to disconnect, or does the remote device wait until the local device physically disconnects before applying the queued GATT operations?

beaufortfrancois:
For DFU case, the GATT operation (characteristic write) causes the remote device to disconnect as you can read at http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk52.v0.9.2%2Fbledfu_appswitching.html&cp=4_0_2_4_2_2_2,

[...]

PR #219, opened 2016-02-28
Also fire `gattserverdisconnected` when a page disconnects itself.

jyasskin:
As discussed in #215.

@thegecko


thegecko:
ack, nice.

beaufortfrancois:
I'll update https://googlechrome.github.io/samples/web-bluetooth/device-disconnect.html as well to remove the `Page is no longer visible...` message then

scheib:
LGTM, with possible additional suggestion of explicitly mentioning that a duplicate event must not arrive when the device is disconnected by the OS.

thegecko:
+1

jyasskin:
@scheib Thanks, done.

PR #228, opened 2016-04-14
Integrate with the Permissions API.

### LINKED ISSUE #164: Sites should be able to find devices they were granted access to previously
If a user pairs a device with an origin through the `requestDevice()` dialog, then restarts the page, they shouldn't need to re-select the device.

Sketch at https://docs.google.com/document/d/1QLEi4lK4St3tS_EQEKqhTOn5Gi47sN5mUoLCm4tOyr8/edit?usp=sharing.

jyasskin:
Demo at https://rawgit.com/jyasskin/web-bluetooth-1/permission-api/index.html. @raymeskhoury @mounirlamouri 

This exclusively uses the Permissions API's structures to record whether
a site has permission to use a device. These will get updated better as
the platform changes/refines its permission design.

The Permissions public algorithms don't handle revocation quite right
yet, so this patch calls into some of the private ones.

Fixes #164.

@annevk, I'm thinking of the [monkeypatching](https://rawgit.com/jyasskin/web-bluetooth-1/permission-api/index.html#permission-api-integration) as temporary while the [definition of "bluetooth"](https://rawgit.com/jyasskin/web-bluetooth-1/permission-api/index.html#dom-permissionname-bluetooth) makes it through Shepherd, and then I'd update [PermissionName](https://w3c.github.io/permissions/#enumdef-permissionname) to link back to here. Thoughts?


annevk:
Seems reasonable I suppose. That's the preferred way to integrate with the Permissions API?

jyasskin:
It's my current guess at the right way to integrate with the Permissions API. We'll see how it goes, and maybe correct later.

index.bs
```diff
@@ -976,17 +1061,20 @@ spec: html
...
       <a lt="add an allowed bluetooth device"
-         >Add <var>device</var> to the origin's allowed devices map.</a>
+      >Add <var>device</var> to <var>storage</var></a>
       with the union of <var>requiredServiceUUIDs</var>
       and <var>optionalServiceUUIDs</var> as <var>allowed services</var>.
     </li>
     <li>
       <a>Get the <code>BluetoothDevice</code> representing</a> <var>device</var>
-      inside the <a>context object</a>
+      inside the <a>context object</a> and <var>storage</var>
```
g-ortuno: Don't you need to add the device to `status.devices`?

index.bs
```diff
@@ -1096,6 +1184,186 @@ spec: html
...
+          <dfn attribute for="AllowedBluetoothDevice">\[[device]]</dfn>
+          that holds a <a>Bluetooth device</a>.
+        </p>
+      </dd>
+      <dt>
+        <a>permission result type</a>
+      </dt>
+      <dd>
+        <pre class="idl">
+          interface BluetoothPermissionResult : PermissionStatus {
```
g-ortuno: I think it's confusing that this type is called `Result` but it inherits from `Status`. Is that how it's usually done?

index.bs
jyasskin: Oops, yes.

index.bs
jyasskin: This is the first subclass of `PermissionStatus`, so there's no "usually" yet. I'm worried that `BluetoothPermissionStatus` would imply that it reflects the status of the entire permission (which it doesn't because it tends to return just the devices relevant to a particular query or request). I tried to [rename `PermissionStatus` to `PermissionResult`](https://github.com/w3c/permissions/pull/82), but because it's already shipped, @slightlyoff suggested that we should just name subclasses "...Result" and leave the base class alone.

g-ortuno: Ah OK.

g-ortuno:
lgtm

Issue #163, opened 2015-09-13
When a device causes a page to open, the page should be able to find the device


jyasskin:
For example, an [Eddystone-URL](https://github.com/google/eddystone/tree/master/eddystone-url) beacon can cause the browser to suggest a user visit a particular web page. If the user does, that page could get direct access to the device instead of needing to go through the `requestDevice()` chooser.

Sketch at https://docs.google.com/document/d/1jFCTyq84T2fLc8ZhxorTz3u_gCk59hp17EmkFgaDQ2c/edit?usp=sharing.

mfoltzgoogle:
Thanks for the document @jyasskin, that explains pretty well what I was thinking of regarding the question I asked at yesterday's F2F.

FWIW I like the approach using `navigator.bluetooth.referringDevice`, as it makes the page initialization code pretty clean.  Also things appended to the URL might get lost or munged during redirects.

```
if (navigator.bluetooth.referringDevice) {
  connectToServices(navigator.bluetooth.referringDevice);
} else {
  connectDeviceButton.onclick = function() { navigator.bluetooth.requestDevice
      .then(connectToServices); };
}
```

jyasskin:
Before we can do this, we need to ensure it doesn't make it really easy for a hostile physical-web device to get paired as a keyboard: https://github.com/google/physical-web/issues/568. Thanks @hillbrad for pointing out the problem during TPAC.

beaufortfrancois:
If we go down the Permissions API road (https://lists.w3.org/Archives/Public/public-web-bluetooth/2015Dec/0007.html), it would be as simple as adding a `referringDevice` key to the Permission object.

```js       
permissions.query({ name: 'bluetooth' })
.then(permission => {
  if (permission.state != 'granted') {
    console.log('User has forbidden websites to use bluetooth...');
    return;
  }
  if (permission.referringDevice) {
    // Let's try to connect to the Physical Web object first. 
    return connectDevice(permission.referringDevice);
  }
  ...

PR #229, opened 2016-04-20
Remove the device info and device class fields from BluetoothDevice.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/remove-device-info/index.html#bluetoothdevice.

These fields are in the API because it started life as [chrome.bluetooth.Device](https://developer.chrome.com/apps/bluetooth#type-Device), but they don't exist in either the [Core Bluetooth](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBPeripheral_Class/index.html) or (except for device class) the [Android](http://developer.android.com/reference/android/bluetooth/BluetoothDevice.html#lfields) APIs. Developers can still navigate the GATT tree to look them up; they just won't be available at the top level anymore.

I asked for opinions on doing this at
https://lists.w3.org/Archives/Public/public-web-bluetooth/2016Apr/0017.html
and the limited response was positive.

This change also makes the "get the BluetoothDevice representing a
Bluetooth device" algorithm synchronous, which will help implement #163.

@beaufortfrancois FYI, we'll have to rewrite the https://googlechrome.github.io/samples/web-bluetooth/device-info.html once this makes it to Chromium.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -38,12 +38,6 @@ Markup Shorthands: css no, markdown yes
     "status": "Living Standard",
     "publisher": "Bluetooth SIG"
   },
-  "BLUETOOTH-ASSIGNED-BASEBAND": {
-    "href": "https://www.bluetooth.org/en-us/specification/assigned-numbers/baseband",
-    "title": "Assigned Numbers for Baseband",
-    "status": "Living Standard",
-    "publisher": "Bluetooth SIG"
-  },
   "BLUETOOTH-ASSIGNED-SERVICES": {
     "href": "https://developer.bluetooth.org/gatt/services/Pages/ServicesHome.aspx",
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
I'll all for this PR.
I'll update https://googlechrome.github.io/samples/web-bluetooth/device-info.html right away as there were no need for this.

beaufortfrancois:
WIP at https://github.com/GoogleChrome/samples/pull/324

beaufortfrancois:
https://googlechrome.github.io/samples/web-bluetooth/device-info.html has been updated.

beaufortfrancois:
FYI. Chrome implementation has been updated: https://chromium.googlesource.com/chromium/src/+/10e8933627cf2753881d6106ec9f480454f0d72d

PR #230, opened 2016-04-25
Specify the Eddystone upgrade.

### LINKED ISSUE #163: When a device causes a page to open, the page should be able to find the device

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/eddystone-upgrade/index.html.

Much of the content in this patch is about letting a UA not constrain the set of services a site can access on a device that asked to open it. The threads at https://lists.w3.org/Archives/Public/public-web-bluetooth/2016Apr/0000.html and https://plus.sandbox.google.com/+FrancoisBeaufort/posts/jGBu1kxHbpZ didn't come to any firm conclusions on this point, but I think it makes sense to assume that a site advertised by a device is cooperating with that device, and so the normal purposes of showing the services list don't apply.

Fixes #163.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -108,6 +108,8 @@ spec: html
         text: event handler idl attribute
         text: global object
         text: in parallel
+        text: initialising a new document object
+        text: navigated
         text: perform a microtask checkpoint
         text: queue a task
         text: relevant settings object
@@ -527,6 +529,8 @@ spec: permissions
 
     interface Bluetooth {
       [SecureContext]
+      readonly attribute BluetoothDevice? referringDevice;
+      [SecureContext]
       Promise&lt;BluetoothDevice> requestDevice(RequestDeviceOptions options);
     };
     Bluetooth implements EventTarget;
@@ -535,6 +539,14 @@ spec: permissions
     Bluetooth implements ServiceEventHandlers;
   </pre>
   <div class="note" heading="{{Bluetooth}} members">
+    <p>
+      {{Bluetooth/referringDevice}} gives access to
+      the device from which the user opened this page, if any.
+      For example, an <a href="https://developers.google.com/beacons/eddystone">Eddystone</a> beacon
+      might advertise a URL, which the UA allows the user to open.
+      A {{BluetoothDevice}} representing the beacon would be available
+      through <code>navigator.bluetooth.{{referringDevice}}</code>.
+    </p>
     <p>
       {{Bluetooth/requestDevice(options)}} asks the user
       to grant this origin access to a device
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -793,9 +809,54 @@ spec: permissions
...
+  </p>
+  <p>
+    Some UAs may allow the user
+    to cause a <a>browsing context</a> to <a lt="navigated">navigate</a>
+    in response to a <a>Bluetooth device</a>.
+    <span class="note">
+      For example, if an
+      <a href="https://developers.google.com/beacons/eddystone">Eddystone</a> beacon
+      advertises a URL,
+      the UA may allow the user to navigate to this URL.
```
kenchris: ...directly ?

index.bs
```diff
@@ -793,9 +809,54 @@ spec: permissions
...
+    </span>
+    If this happens, then as part of <a>initialising a new `Document` object</a>,
```
kenchris: isn't this 'initializing' in American English? or are we not using American English spelling?

index.bs
```diff
@@ -1483,7 +1546,10 @@ spec: permissions
         If one is found, do the following sub-steps.
         <ol>
           <li>
-            Set <code><var>allowedDevice</var>.{{allowedServices}}</code> to
+            If <code><var>allowedDevice</var>.{{allowedServices}}</code> is `"all"`
+            or the all-services-allowed flag is set,
```
kenchris: Shouldn't the flag be marked as <code> or similar?

index.bs
jyasskin: I'm not quite sure what you're asking. The current physical web design in Chrome, that you can see at https://youtu.be/-Y77cUI_z30, just gives the user a link they can click on to navigate to that URL.

index.bs
jyasskin: Heh, that's just the spelling used at https://html.spec.whatwg.org/multipage/browsers.html#initialise-the-document-object. @annevk, is this worth "fixing"?

index.bs
jyasskin: https://html.spec.whatwg.org/multipage/browsers.html#traverse-the-history uses italics, so I'll do that.

index.bs
annevk: See https://github.com/whatwg/html/issues/654. I think it should be fixed, but it's quite a bit of work.

index.bs
kenchris: I re-read it, it is fine :-)

beaufortfrancois:
LGTM

index.bs
```diff
@@ -779,9 +791,57 @@ spec: permissions
...
+    </li>
+    <li>
+      <a>Get the <code>BluetoothDevice</code> representing</a> <var>referringDevice</var>
+      inside `navigator.bluetooth`,
+      and let |referringDeviceObj| be the result.
+    </li>
+    <li>
+      If the previous step threw an exception, abort these steps.
+
+      Note: This means the UA didn't infer that the user intended to
```
scheib: This seems fine spec wise, but understandability wise I don't have an idea come to mind for _why_ the UA wouldn't infer access.

index.bs
```diff
@@ -1587,7 +1648,8 @@ spec: permissions
         <td><dfn>\[[allowedServices]]</dfn></td>
         <td>`undefined`</td>
         <td>
-          This device's {{AllowedBluetoothDevice/allowedServices}} list for this origin.
+          This device's {{AllowedBluetoothDevice/allowedServices}} list for this origin
+          or `"all"` if all services are allowed.
```
scheib: An inline note here explaining when "all" might be expected (referringDevice) would help. e.g. 'E.G. When a referringDevice provides access to a device the User Agent may infer that access to all services on the device is appropriate. Selecting a limited set of allowedServices would require a way for that set to be communicated from the referringDevice."

scheib:
LGTM, optional suggestions for additional notes.

index.bs
jyasskin: "For example, the user might have denied GATT access globally."

index.bs
jyasskin: "For example, a UA may grant an origin access to all services on a {{referringDevice}} that advertised a URL on that origin."

PR #233, opened 2016-04-27
Use the [SecureContext] IDL attribute to enforce secure contexts.

jyasskin:
Instead of manually throwing an exception from requestDevice().

@mikewest, is [`[SecureContext]`](https://heycam.github.io/webidl/#SecureContext) ready to use here? Do you recommend putting it on just the entry function, or on all the interfaces? (It's also not implemented in Chrome yet, but I assume that's just a matter of time.)

Preview at https://rawgit.com/jyasskin/web-bluetooth-1/secure-context-attribute/index.html#device-discovery

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,7 +111,6 @@ spec: html
         text: event handler idl attribute
         text: global object
         text: in parallel
-        text: incumbent settings object
         text: perform a microtask checkpoint
         text: queue a task
         text: relevant settings object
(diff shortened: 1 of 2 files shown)


jyasskin:
Ping @mikewest.

jyasskin:
Via IM, Mike thinks things are ready, and that it's reasonable to only block the entry points and not the interfaces.

beaufortfrancois:
For info, https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/ILe65HVid88

PR #235, opened 2016-05-03
Have sites explicitly watch for advertisements on paired devices.

### LINKED ISSUE #117: Describe what happens when an advertising packet is received.
We need to at least update [`BluetoothDevice.adData`](http://webbluetoothcg.github.io/web-bluetooth/#idl-def-bluetoothdevice-addata).

### LINKED ISSUE #212: Redesign access to advertising data for a single device.
Current spec for access to the advertising data of a single device appears as:

```
interface BluetoothDevice {
  readonly attribute BluetoothAdvertisingData adData;
  ...
}
interface BluetoothAdvertisingData {
  readonly attribute unsigned short? appearance;
  readonly attribute byte? txPower;
  readonly attribute byte? rssi;
  readonly attribute BluetoothManufacturerDataMap manufacturerData;
  readonly attribute BluetoothServiceDataMap serviceData;
};
```

However, as discussed on https://code.google.com/p/chromium/issues/detail?id=568137 (ortuno & jyasskin) this is limited in many ways:

- You have to be actively scanning to get the TxPower and the RSSI
- Many devices don't advertise if they are connected to.
- [comment 6](https://code.google.com/p/chromium/issues/detail?id=568137#c6) details on RSII updates from Adv packets and connected device polling.

> We could safely let web sites "scan" for devices they already have access to, which would let them opt into the increased power use and get advertising data updates. I think we'd change the .adData field into a "watch" function and event. Thoughts?

This issue is distinct from issue [Let websites scan for nearby advertisements](https://github.com/WebBluetoothCG/web-bluetooth/issues/191) which would scan for all advertisements. Though, perhaps if #191 is solved this issue may be omitted as a subset of what could be done scanning for all AD packets.

jyasskin:
This removes the enforcement that a site can only see service data for services
from its allowed services list. Since an advertisement is broadcast and
read-only, I think there's neither any private data in the advertisement nor a
need to protect the device from hostile sites.

Fixes #117 and fixes #212.

Preview at https://rawgit.com/jyasskin/web-bluetooth-1/watch-advertisements/index.html#advertising-events.


beaufortfrancois:
Thank you for spec'ing this @jyasskin!

Here are my two cents: Do we really need the added this new boolean attribute `watchingAdvertisements`? Can't we simply set it internally to true when developer calls `device.addEventListener('advertisementreceived', foo);`?

It would be as simple and nice as:
```js
var known_service = "A service in the iBeacon’s GATT server";
return navigator.bluetooth.requestDevice({
  filters: [{services: [known_service]}]
}).then(device => {
  // Set `watchingAdvertisements` to true
  device.addEventListener('advertisementreceived', interpretIBeacon);
});

// Later on...
// Set `watchingAdvertisements` to false
device.removeEventListener('advertisementreceived', interpretIBeacon);

```

beaufortfrancois:
If we really want these two steps, I'd advise for consistency with `startNotifications()/stopNotifications()` and make `watchingAdvertisements` a function like `startWatchingAdvertisements()/stopWatchingAdvertisements()`:

```js
var known_service = "A service in the iBeacon’s GATT server";
return navigator.bluetooth.requestDevice({
  filters: [{services: [known_service]}]
}).then(device => {
  device.startWatchingAdvertisements();
  device.addEventListener('advertisementreceived', interpretIBeacon);
});

// Later on...
device.stopWatchingAdvertisements();
```

jefBinomed:
I shared the point of view of @beaufortfrancois with the fact that the device.watchingAdvertisement is useless.

I you look at other sensors, you subscribe the app to an event if you want to receive it and remove the listener when you want to continue to listen to it... So If i could vote, i will vote for option 1 of francois

index.bs
```diff
@@ -1611,9 +1589,12 @@ spec: permissions
       interface BluetoothDevice {
         readonly attribute DOMString id;
         readonly attribute DOMString? name;
-        readonly attribute BluetoothAdvertisingData adData;
         readonly attribute BluetoothRemoteGATTServer gatt;
         readonly attribute FrozenArray&lt;UUID> uuids;
+
+        void watchAdvertisements();
```
jyasskin: Maybe @beaufortfrancois's `startWatchingAdvertisements()` is better since it avoids the nonword "unwatch".

index.bs
```diff
@@ -1833,17 +1855,42 @@ spec: permissions
         interface BluetoothServiceDataMap {
           readonly maplike&lt;UUID, DataView>;
         };
-        interface BluetoothAdvertisingData {
+        [Constructor(DOMString type, BluetoothAdvertisingEventInit init)]
+        interface BluetoothAdvertisingEvent : Event {
+          readonly attribute BluetoothDevice device;
```
jyasskin: This could just be the `target` of the Event, since these are always dispatched to a `BluetoothDevice`. However, "target" is a little weird, since it's actually the *source* of the advertisement.

index.bs
g-ortuno: Hmm I was expecting this to be a promise. The main reason is that starting to scan could fail for example if the adapter is off.

index.bs
```diff
@@ -1909,56 +1962,179 @@ spec: permissions
...
+                <dd>Append the listed UUIDs to <code><var>event</var>.uuids</code>.</dd>
+
+                <dt>Shortened <a lt="Local Name Data Type">Local Name</a></dt>
+                <dt>Complete <a lt="Local Name Data Type">Local Name</a></dt>
+                <dd>
+                  <a>UTF-8 decode without BOM</a> the AD data and
+                  set <code><var>event</var>.name</code> to the result.
+
+                  Note: We don't expose whether the name is complete
+                  because Android and Core Bluetooth don't.
```
g-ortuno: Actually I think we can get this from Android because it exposes the raw bytes: http://developer.android.com/reference/android/bluetooth/le/ScanRecord.html#getBytes()

index.bs
jyasskin: Mhmm. I was thinking that we'd remember that a scan was supposed to be happening, and start it when the adapter comes on. [Android](http://developer.android.com/reference/android/bluetooth/le/BluetoothLeScanner.html#startScan(java.util.List<android.bluetooth.le.ScanFilter>, android.bluetooth.le.ScanSettings, android.bluetooth.le.ScanCallback)) can fail asynchronously, but it doesn't report success, so we can't use that to resolve a Promise, and [CoreBluetooth](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/index.html#//apple_ref/occ/instm/CBCentralManager/scanForPeripheralsWithServices:options:) doesn't fail at all. It looks like [Windows 10](https://msdn.microsoft.com/en-us/library/windows/apps/windows.devices.bluetooth.advertisement.bluetoothleadvertisementwatcher.start.aspx) does report both success and failure synchronously, and has a reasonable [list of errors](https://msdn.microsoft.com/en-us/library/windows/apps/windows.devices.bluetooth.bluetootherror.aspx).

Ok, let's make this a Promise, and just have a bunch of platforms never fail.

index.bs
jyasskin: Yeah, it's more about following precedent. If most users don't need this (shown by the existing platforms not providing it), we shouldn't provide it? If we do provide it, I think it should be a boolean rather than two distinct fields. My comment could be better though.

g-ortuno: Yeah, I'm OK with not providing it, just wanted to correct the `Note`.

g-ortuno:
lgtm!

jyasskin:
As discussed in https://github.com/WebBluetoothCG/web-bluetooth/issues/236#issuecomment-216833443, we can't make `addEventListener` have side-effects, so it has to be an attribute or a function.

Emill:
What's really the point of the public `watchingAdvertisements` property?
If you know the promise has resolved then it's watching until you call unwatchAdvertisements().

Issue #234, opened 2016-05-02
Allow requestDevice for all devices


scheib:
Initially the Web Bluetooth specification has required requestDevice options to filter the set of devices that will be displayed for the user to select from. One argument against allowing developers to request all devices be shown is that out of laziness they may often do so to the detriment of the user's experience.

When devices do not advertise their services and do not have a reliable name that can be used as a filter then there is no way to select devices.

One concrete example is this [Bluetooth Rename Tool](https://rawgit.com/scheib/webbluetoothcg-demos/bluetooth-rename/bluetooth-rename/index.html) I made recently (link will break eventually, maybe it will appear on [demos](https://github.com/WebBluetoothCG/demos) page eventually). This tool uses `generic_access`'s `gap.device_name`, however `generic_access` is prohibited from being advertized. It is unknown what device names may be already that the user wishes to edit.

I believe other developers will eventually encounter this for devices that don't advertise services and have a multitude of names. Though this should be a small fraction of use.

This demo currently works in Chrome due to a bug allowing a filter of `name: ""` to return all devices. That isn't the right way to solve this. I'm not clear on how it should be solved, but would like to communicate to developers that they should only ask for all devices when they really need to. Only half jokingly:

```js
navigator.bluetooth.requestDevice({
  filters: [{
    'listAllDevicesEvenThoughItIsAPoorUserExperience': true
  }],
  optionalServices: ['generic_access']
})
```

nemik:
I can definitely see a need for limiting a scan-everything mode, but in that case perhaps the filters should be expanded a little bit. Sorry, I'm new to the community here so maybe I haven't seen all the documentation and specs, but according to: https://webbluetoothcg.github.io/web-bluetooth/#dom-requestdeviceoptions-filters the only filters are for service-UUID's and names. Having one for manufacturer UUID would be nice.

One implementation I like very much, on Android, is: http://altbeacon.github.io/android-beacon-library/javadoc/org/altbeacon/beacon/BeaconParser.html
I understand that can be too much to pack into a web-spec, but some way to specify flexible filtering would be nice.

For my purpose, I have to pack my sensor data and other info into an advertisement. It would be nice to not need to use additional, precious, advertising bytes on things like service UUID's or names.

youcangetme:
OK, so the Web Bluetooth spec is all about connecting devices around me to a website so that I can interact with that website in more interactive and complex fashion, right? What about apps for artists or musicians?

[...]

PR #239, opened 2016-05-09
Specify a Bluetooth Scanning API.

### LINKED ISSUE #191: Let websites scan for nearby advertisements.
Websites should be able to register for an event when new BLE advertising data is received, even if they don't have permission to send information to that device. This reveals the user's location, so we'll need a permission grant, and we may want to let users set bounds on the area or time in which the grant applies. We need to figure out if a site with full geolocation access should automatically get access to scan for beacons.

We also need to figure out whether it's useful to constrain scanning to tabs, or whether we need to figure out ServiceWorker access from the start. From the use case of walking around a museum listening to the sound corresponding to the closest exhibit, I'm inclined to think we have to figure out full background access from the start.

jyasskin:
This doesn't cover scanning from a service worker yet.

Fixes #191.

Previews at https://rawgit.com/jyasskin/web-bluetooth-1/scanning/scanning.html and https://rawgit.com/jyasskin/web-bluetooth-1/scanning/index.html#bluetoothdevice

### DIFF
--- a/scanning.bs
+++ b/scanning.bs
@@ -269,7 +269,6 @@ spec: html
       BluetoothServiceUUID serviceUUID;
       unsigned short manufacturerData;
       BluetoothServiceUUID serviceData;
-      boolean connectable;
     };
 
     dictionary BluetoothLEScanOptions {
@@ -421,7 +420,6 @@ spec: html
         readonly attribute UUID? serviceUUID;
         readonly attribute unsigned short? manufacturerData;
         readonly attribute UUID? serviceData;
-        readonly attribute boolean? connectable;
       };
 
       interface BluetoothLEScan {
@@ -700,16 +698,6 @@ spec: html
           If <code>|filter|.serviceData</code> is non-`null`,
           it is the 128-bit UUID representation of the UUID
           in some <a>Service Data</a> in |event|.
-        </li>
-        <li>
-          If <code>|filter|.connectable</code> is `true`,
-          |event| has either the <a>ADV_IND</a>
-          or the <a>ADV_DIRECT_IND</a> type.
-        </li>
-        <li>
-          If <code>|filter|.connectable</code> is `false`,
-          |event| has either the <a>ADV_NONCONN_IND</a>
-          or the <a>ADV_SCAN_IND</a> type.
         </li>
       </ul>
     </div>
--- a/scanning.bs
+++ b/scanning.bs
@@ -71,54 +71,16 @@ Markup Shorthands: css no, markdown yes
 </pre>
 
 <pre class="anchors">
-spec: BLUETOOTH-ASSIGNED
-    type: enum; urlPrefix: https://developer.bluetooth.org/gatt/
-        urlPrefix: characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.
-            text: org.bluetooth.characteristic.body_sensor_location; url: body_sensor_location.xml#
-            text: org.bluetooth.characteristic.gap.appearance; url: gap.appearance.xml#
-            text: org.bluetooth.characteristic.heart_rate_control_point; url: heart_rate_control_point.xml#
-            text: org.bluetooth.characteristic.heart_rate_measurement; url: heart_rate_measurement.xml#
-            text: org.bluetooth.characteristic.ieee_11073-20601_regulatory_certification_data_list; url: ieee_11073-20601_regulatory_certification_data_list.xml#
-            text: org.bluetooth.characteristic.pnp_id; url: pnp_id.xml#
-        urlPrefix: descriptors/Pages/DescriptorViewer.aspx?u=org.bluetooth.descriptor.
-            text: org.bluetooth.descriptor.gatt.characteristic_presentation_format; url: gatt.characteristic_presentation_format.xml#
-            text: org.bluetooth.descriptor.gatt.client_characteristic_configuration; url: gatt.client_characteristic_configuration.xml#
-        urlPrefix: services/Pages/ServiceViewer.aspx?u=org.bluetooth.service.
-            text: org.bluetooth.service.cycling_power; url: cycling_power.xml#
-            text: org.bluetooth.service.device_information; url: device_information.xml#
-            text: org.bluetooth.service.heart_rate; url: heart_rate.xml#
-    type: dfn
-        text: Shortened Local Name; url: https://www.bluetooth.org/en-us/specification/assigned-numbers/generic-access-profile#
-
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
-    type: dfn
-        text: Realm; url: sec-code-realms
     type: method
         text: Array.prototype.map; url: sec-array.prototype.map
     type: interface
-        text: Array; url: sec-array-objects
-        text: ArrayBuffer; url: sec-arraybuffer-constructor
-        text: DataView; url: sec-dataview-constructor
-        text: Map; url: sec-map-constructor
-        text: Promise; url:sec-promise-objects
-        text: Set; url: sec-set-objects
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
-        text: TypedArray; url: sec-typedarray-constructors
-
-spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/#
-    type: dfn
-        text: a copy of the bytes held; url: dfn-get-buffer-source-copy
 </pre>
 <pre class="link-defaults">
 spec: html
     type: dfn
-        text: browsing context
-        text: environment settings object
-        text: event handler idl attribute
-        text: global object
         text: in parallel
-        text: incumbent settings object
-        text: perform a microtask checkpoint
         text: queue a task
         text: relevant settings object
         text: responsible event loop
@@ -266,6 +228,8 @@ spec: html
 
   <pre class="idl">
     dictionary BluetoothLEScanFilterInit {
+      DOMString name;
+      DOMString namePrefix;
       BluetoothServiceUUID serviceUUID;
       unsigned short manufacturerId;
       BluetoothServiceUUID serviceDataUUID;
@@ -417,6 +381,8 @@ spec: html
     <pre class="idl">
       [Constructor(BluetoothLEScanFilterInit init)]
       interface BluetoothLEScanFilter {
+        readonly attribute DOMString? name;
+        readonly attribute DOMString? namePrefix;
         readonly attribute UUID? serviceUUID;
         readonly attribute unsigned short? manufacturerId;
         readonly attribute UUID? serviceDataUUID;
@@ -686,6 +652,18 @@ spec: html
       </p>
       <ul>
         <li>
+          If <code>|filter|.name</code> is non-`null`,
+          |event| has a <a>Local Name</a> equal to <code>|filter|.name</code>.
+
+          Note: A <a>Shortened Local Name</a>
+          can match a {{BluetoothLEScanFilter/name}} filter.
+        </li>
+        <li>
+          If <code>|filter|.namePrefix</code> is non-`null`,
+          |event| has a <a>Local Name</a>,
+          and <code>|filter|.namePrefix</code> is a prefix of it.
+        </li>
+        <li>
           If <code>|filter|.serviceUUID</code> is non-`null`,
           it is equal to a <a>Service UUID</a> in |event|.
         </li>
@@ -872,6 +850,11 @@ spec: html
             <li value="1">Data Types Definitions and Formats
               <ol>
                 <li value="1"><dfn>Service UUID</dfn></li>
+                <li value="1">
+                  <dfn>Local Name</dfn>
+                  also defines <dfn>Shortened Local Name</dfn>
+                  and <dfn>Complete Local Name</dfn>
+                </li>
                 <li value="4"><dfn>Manufacturer Specific Data</dfn></li>
                 <li value="11"><dfn>Service Data</dfn></li>
               </ol>
--- a/scanning.bs
+++ b/scanning.bs
@@ -273,8 +273,9 @@ spec: html
     };
 
     dictionary BluetoothLEScanOptions {
-      required sequence&lt;BluetoothLEScanFilterInit> filters;
+      sequence&lt;BluetoothLEScanFilterInit> filters;
       boolean keepRepeatedDevices = false;
+      boolean acceptAllAdvertisements = false;
     };
 
     partial interface Bluetooth {
@@ -311,6 +312,10 @@ spec: html
       doesn't guarantee you won't get redundant events;
       it just allows the UA to save power by omitting them.
     </p>
+    <p>
+      In the rare case that you want to receive every advertisement without filtering them,
+      use the <dfn dict-member for="BluetoothLEScanOptions">acceptAllAdvertisements</dfn> field.
+    </p>
   </div>
 
   <div algorithm>
@@ -320,6 +325,20 @@ spec: html
       and run the following steps <a>in parallel</a>:
     </p>
     <ol class="algorithm">
+      <li>
+        If <code>|options|.acceptAllAdvertisements</code> is `true`,
+        and <code>|options|.filters</code> is present,
+        <a>reject</a> |promise| with a {{TypeError}} and abort these steps.
+
+        Note: There's no need to include filters if all advertisements are being accepted.
+      </li>
+      <li>
+        If <code>|options|.acceptAllAdvertisements</code> is `false`,
+        and <code>|options|.filters</code> is either absent or empty,
+        <a>reject</a> |promise| with a {{TypeError}} and abort these steps.
+
+        Note: An empty set of filters wouldn't return any advertisements.
+      </li>
       <li>
         Let |filters| be
         <code>|options|.filters.map(filter=>new
@@ -334,12 +353,11 @@ spec: html
             name: <a idl>"bluetooth-le-scan"</a>,
             filters: <var>options</var>.filters,
             keepRepeatedDevices: <var>options</var>.keepRepeatedDevices,
+            acceptAllAdvertisements: <var>options</var>.acceptAllAdvertisements,
           }
         </pre>
 
-        <p class="note">
-          This may require that this algorithm was <a>triggered by user activation</a>.
-        </p>
+        Note: This may require that this algorithm was <a>triggered by user activation</a>.
       </li>
       <li>
         If the result is {{"denied"}},
@@ -354,6 +372,10 @@ spec: html
           <tr>
             <td>{{BluetoothLEScan/keepRepeatedDevices}}</td>
             <td><code>|options|.keepRepeatedDevices</code></td>
+          </tr>
+          <tr>
+            <td>{{BluetoothLEScan/acceptAllAdvertisements}}</td>
+            <td><code>|options|.acceptAllAdvertisements</code></td>
           </tr>
           <tr><td>{{BluetoothLEScan/active}}</td><td>`true`</td></tr>
         </table>
@@ -405,6 +427,7 @@ spec: html
       interface BluetoothLEScan {
         readonly attribute FrozenArray&lt;BluetoothLEScanFilter> filters;
         readonly attribute boolean keepRepeatedDevices;
+        readonly attribute boolean acceptAllAdvertisements;
 
         readonly attribute boolean active;
 
@@ -428,6 +451,12 @@ spec: html
       <ol class="algorithm">
         <li>
           Initialize all fields to `null`.
+        </li>
+        <li>
+          If no member of |init| is present, throw a {{TypeError}}.
+
+          Note: A filter can't implicitly allow all advertisements.
+          Use {{BluetoothLEScanOptions/acceptAllAdvertisements}} to explicitly do it.
         </li>
         <li>
           For each <a>present</a> member in |init|,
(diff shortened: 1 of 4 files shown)


scanning.bs
```diff
@@ -0,0 +1,993 @@
...
+
+    <div algorithm="bluetoothlescan-match">
+    <p>
+      An advertising event |event|
+      <dfn for="BluetoothLEScan" lt="match">matches</dfn> a {{BluetoothLEScan}} |scan|
+      if the following steps return `match`:
+    </p>
+    <ol>
+      <li>
+        If |event| doesn't <a for="BluetoothLEScanFilter">match</a>
```
jyasskin: Note that this leaves no way to do a completely open scan. It has to be filtered for some UUID or Manufacturer ID. Is that OK, or do we need an extra filter type to match everything? (I don't want to make no-filters the sign for completely-open because that makes the API non-monotonic: reducing the number of filters would match less, less, less, and then more advertisements.)

scheib: Similar to #234 I think developers will have good use cases for matching everything.

beaufortfrancois:
I wonder why developers would use `requestDevice` rather than `requestLEScan`. The latter seems more powerful and better in terms of UX. Could you comment on the drawbacks?

jyasskin:
@beaufortfrancois At the moment, `requestLEScan` gets you a bunch of `BluetoothDevice` objects, but all of their `.gatt` fields are `null` (unless you already had permission to communicate with them), so you have to use `requestDevice` to communicate. I think it's worthwhile to give folks a way to upgrade, but we need to ask the security UI folks what it should look like, and it'll probably be designed to push people toward just using `requestDevice()` if they don't really need to control the scan.

beaufortfrancois:
This makes more sense now ;)
Thank you @jyasskin!

So once you have a `BluetoothDevice` object, the way to know if you have to use `requestDevice` is to check `device.gatt`? But how do you make sure that only this device will be shown in the bluetooth chooser?

```js

navigator.bluetooth.requestLEScan({
  filters: [{manufacturerData: 0x004C}],
}).then(() => {
  navigator.bluetooth.addEventListener('advertisementreceived', event => {
    if (!event.device.gatt) {
      // How do I connect to this device specifically?
    }
  })
})

**Comment by jyasskin** [comment]
There's no API for that yet, and I've mailed the security UI folks to ask their opinion on it. The benefits of the chooser are that 1) the user has to pay enough attention to make a choice, and 2) the user can override the page's idea of which device is best. If we ensure that exactly one device appears in the chooser that opens as a result of an advertisement, then we lose both of those. On the other hand, we might be able to add a `hint` field to [`RequestDeviceOptions`](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-requestdeviceoptions), which could ensure that the preferred device is listed first if it matches the other filters.

**Comment by Vudentz** [comment]
Some devices may use non-connectable flag, e.g. broadcaster, in that case requestDevice might actually not work in the end so perhaps we need some field to indicate this to the application.

**Comment by beaufortfrancois** [comment]
The `connectable` field would be indeed valuable to add to the `BluetoothDevice` object. Thanks @Vudentz!
I wonder however if that would be as simple as a boolean field. THere are actually two kinds of connectable there: ADV_IND and ADV_DIRECT_IND
@jyasskin Any thoughts?

**Comment by jyasskin** [comment]
I think we don't need to distinguish between ADV_IND and ADV_DIRECT_IND: the difference is whether the controller can send a SCAN_REQ in response, but the advertising event should just combine data from the advertisement with the SCAN_RSP, so websites don't need to react to the difference.

Do we want to mark this in the advertising event or on the device? I'd initially thought the advertising event, but since we always have to wait for a subsequent advertisement in order to connect, it probably makes sense to turn it to 'true' on the device as soon as we see any connectable advertisement from that device.

File: scanning.bs [hunk 2]
```diff
@@ -0,0 +1,993 @@
...
+        a website could use code like the following:
+      </p>
+      <pre highlight="js">
+        let nearbyBeacons = new Map();
+        let myBeaconUuid = new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
+
+        navigator.bluetooth.<a idl for="Bluetooth" lt="requestLEScan()">requestLEScan</a>({
+          filters: [{manufacturerData: 0x004C}],
+          options: {
+            keepDuplicates: true,
```
beaufortfrancois: `s/keepDuplicates/keepRepeatedDevices/`

jyasskin: Done, thanks.

scanning.bs
jyasskin: This is [`BluetoothLEScanOptions.acceptAllAdvertisements`](https://rawgit.com/jyasskin/web-bluetooth-1/scanning/scanning.html#dom-bluetoothlescanoptions-acceptalladvertisements) now.

Vudentz:
Im quite concerned with these APIs, it is quite low level which may lead us to need to expose scanning intervals which depending on the duty cycle can block other traffic such as GATT to happen. Im also not quite sure how to interpret the aggregation of scan response, there could be other advertisements in the meantime shall the stacks hold the results until it gets a scan response, for requestDevice that may be fine because the API don't except individual reports just a device, but here the holding of events may change the sequence. Perhaps if this is done via device object, so the scan response wouldn't matter and only advertisement would actually cause events.

krantik:
How about timeout? Is the call expected to constantly scan or stop after sometime? Some of the devices(iOS for example) doesn't constantly scan and backs off after some period. Is the same expected here?

krantik:
Is the event expected to have parsed data with separate values for name etc., or raw data and JS side to interpret it?

jyasskin:
@Vudentz 
> Im quite concerned with these APIs, it is quite low level which may lead us to need to expose scanning intervals which depending on the duty cycle can block other traffic such as GATT to happen.

In order to compete with [Android](https://developer.android.com/reference/android/bluetooth/le/ScanCallback.html#onScanResult(int, android.bluetooth.le.ScanResult)) and [Mac](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/index.html#//apple_ref/occ/instm/CBCentralManager/scanForPeripheralsWithServices:options:), I think BlueZ is going to have to report information with this much detail. We can limit the load websites are allowed to put on the system, but beacon use cases definitely need to enable duplicate events, and seem to need the timing of the last advertisement in order to know when things are stale.

If BlueZ doesn't start reporting advertising events, we can infer some of this from changes in the Device object, but it's going to work less well than other platforms.

>Im also not quite sure how to interpret the aggregation of scan response, there could be other advertisements in the meantime shall the stacks hold the results until it gets a scan response, for requestDevice that may be fine because the API don't except individual reports just a device, but here the holding of events may change the sequence. Perhaps if this is done via device object, so the scan response wouldn't matter and only advertisement would actually cause events.

It seems ok to delay the event until after the SCAN_RSP, although maybe we'll find use cases that prefer to get both events. The uses will have to be ok with whatever Android and Mac do, since it's harder to change them.

@krantik 
> How about timeout? Is the call expected to constantly scan or stop after sometime? Some of the devices(iOS for example) doesn't constantly scan and backs off after some period. Is the same expected here?

I could let the UA stop a scan if it's been running "too long". Advertisements aren't reliable anyway, so iOS's periodic scan probably already satisfies the spec, but I could add explicit wording to allow that.

I don't think having folks pass a timeout to `requestLEScan()` will help, since iOS will throttle the scan when it wants regardless of whether the site passed its own timeout. I expect folks to use setTimeout(()=>scan.stop(), xxx) to time out their scans.

> Is the event expected to have parsed data with separate values for name etc., or raw data and JS side to interpret it?

Parsed data. See https://webbluetoothcg.github.io/web-bluetooth/#fire-an-advertisementreceived-event for the parsing. We're nervous about exposing the raw data, both because we can't implement that on [Mac](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManagerDelegate_Protocol/index.html#//apple_ref/occ/intfm/CBCentralManagerDelegate/centralManager:didDiscoverPeripheral:advertisementData:RSSI:), and because some of the fields, like the public addresses hold data that we're not sure websites should know.

Vudentz:
@jyasskin 
> In order to compete with Android and Mac, I think BlueZ is going to have to report information with this much detail. We can limit the load websites are allowed to put on the system, but beacon use cases definitely need to enable duplicate events, and seem to need the timing of the last advertisement in order to know when things are stale.

BlueZ is used in very different systems as Android, it is quite common to have multiple applications using the API in parallel, perhaps we could compare it to Mac but looking at its API it seems to be very similar to the kind of access BlueZ offers with StartDiscovery combined with SetDiscoveryFilter. Btw, it is already possibility to enable duplicated reports, in that case the RSSI property is emitted every time we receive a report from the peripheral, similar to what is stated in the Mac documentation.

> It seems ok to delay the event until after the SCAN_RSP, although maybe we'll find use cases that prefer to get both events. The uses will have to be ok with whatever Android and Mac do, since it's harder to change them.

Well we do wait for the scan response so the applications can see the device name, etc, which is quite convenient for requestDevice but it perhaps wouldn't be for requestLEScan, anyway as long as it not strictly forbidden I guess we can live with that.

Btw, we also have to keep in mind there could be peripherals connected which makes requestLEScan even less reliable as it cannot use very high scan duty cycle in such cases.

scanning.bs
```diff
@@ -0,0 +1,896 @@
...
+          in some <a>Manufacturer Specific Data</a> in |event|.
+        </li>
+        <li>
+          If <code>|filter|.serviceData</code> is non-`null`,
+          it is the 128-bit UUID representation of the UUID
+          in some <a>Service Data</a> in |event|.
+        </li>
+        <li>
+          If <code>|filter|.connectable</code> is `true`,
+          |event| has either the <a>ADV_IND</a>
```
jyasskin: @g-ortuno points out that this is unimplementable in the [Android API](https://developer.android.com/reference/android/bluetooth/le/ScanResult.html).

@beaufortfrancois @jracle do you think it's better for this to work inconsistently depending on the platform, or to leave out the connectable bit until Android adds support for it?

nik-markovic:
Hi,
I had a few questions/comments:

1) BluetoothLEScanFilterInit.manufacturerData, shouldn't this be renamed into "manufacturerId"? Similarly serviceData to "serviceDataUUID".

2) @jyasskin I agree that the API should not form such a tight association to "advertisement" packets. I believe that the LE Bluetooth spec does not restrict contents of scan response and advertisement indication. Though likely rare, for example, a device may have manufacturer data in scan response and service data in advertisement indication. I think that the events should be called "scan responses" (as in: response to user's request to scan for devices) or "scan events" in order to avoid this coupling.

3) Shouldn't typical scan filters be applicable to this scan? For example, matching on a name.

4) @jyasskin I don't see mention missing .gatt object on the returned device. Is there any thing else missing? I hope "id" is available, so the user's code can maintain a map of value.  Can't we still expose .gatt object on returned devices and allow the code to initiate a connect request, requiring a user to answer a prompt?

jyasskin:
> 1) BluetoothLEScanFilterInit.manufacturerData, shouldn't this be renamed into "manufacturerId"? Similarly serviceData to "serviceDataUUID".

My thought with those fields was that we could later extend them to whole objects with all the complexity of https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html, so I wanted to use the same names as the AD field. That said, I can see some sense in using a different name for the shorthand. While I mull it over, 👍 this comment to prefer `manufacturerId` and `serviceDataUUID` or 👎 this comment to prefer `manufacturerData` and `serviceData`?

jyasskin:
> 2) @jyasskin I agree that the API should not form such a tight association to "advertisement" packets. I believe that the LE Bluetooth spec does not restrict contents of scan response and advertisement indication. Though likely rare, for example, a device may have manufacturer data in scan response and service data in advertisement indication. I think that the events should be called "scan responses" (as in: response to user's request to scan for devices) or "scan events" in order to avoid this coupling.

I believe the spec consistently refers to [advertising events](https://rawgit.com/jyasskin/web-bluetooth-1/scanning/scanning.html#advertising-events) instead of either advertising packets or scan responses, which comes from BT4.2 1.A.1.2 saying that an advertising event comprises an advertising packet and resulting scan requests and responses, on each of the three advertising channels. So I think this is ok as-is, and that using "scan events" would diverge too much from the Bluetooth spec.

jyasskin:
>3) Shouldn't typical scan filters be applicable to this scan? For example, matching on a name.

I think the `name` and `namePrefix` filters are the only ones I have in [`requestDevice()`](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-bluetoothrequestdevicefilter) but not [`requestLEScan()`](https://rawgit.com/jyasskin/web-bluetooth-1/scanning/scanning.html#dictdef-bluetoothlescanfilterinit). Android provides an [exact name match](https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html#setDeviceName(java.lang.String)), but not a partial name match, which limits the efficiency gains we can get by specifying a partial match. (Of course, iOS has [much weaker filters](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/index.html#//apple_ref/occ/instm/CBCentralManager/scanForPeripheralsWithServices:options:).) My impression of `requestDevice()` uses is that most have needed to use partial name matches.

So, I'm inclined not to include `name` filters in `requestLEScan()` until we find concrete use cases.

jyasskin:
> 4) @jyasskin I don't see mention missing .gatt object on the returned device. Is there any thing else missing? I hope "id" is available, so the user's code can maintain a map of value.

Only [`.gatt` is newly optional on `BluetoothDevice`](https://rawgit.com/jyasskin/web-bluetooth-1/scanning/index.html#bluetoothdevice). `.id` is still available and should be stable until the user clears their cookies or revokes permission. The wording about "[`"bluetooth"`'s extra permission data](https://rawgit.com/jyasskin/web-bluetooth-1/scanning/scanning.html#advertising-events)" covers this while giving UAs enough freedom to reset IDs at the right time.

> Can't we still expose .gatt object on returned devices and allow the code to initiate a connect request, requiring a user to answer a prompt?

Yes, I plan to add a way to get a `BluetoothRemoteGATTServer` from a scanned device in a future revision. I think we have agreement on what that interface should be, but I'd like to do it in a dedicated PR to make sure we get the right attention.

scanning.bs
jracle: @jyasskin my proposal is to add a nullable boolean to BluetoothDevice object ( see #253 ) 

```
interface BluetoothDevice {
  ...
  readonly attribute boolean? connectable;
};
```

I'm frustrated by those platform inconsistencies, but due to them, I'm more comfortable leaving out the connectable filter for now, and adding above proposal instead. In that case, we would have a consistent behavior, checking for the boolean before attempting connection.

nik-markovic:
>> 1) BluetoothLEScanFilterInit.manufacturerData, shouldn't this be renamed into "manufacturerId"? Similarly serviceData to "serviceDataUUID".

>My thought with those fields was that we could later extend them to whole objects with all the complexity of https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html, so I wanted to use the same names as the AD field. That said, I can see some sense in using a different name for the shorthand. While I mull it over, :+1: this comment to prefer manufacturerId and serviceDataUUID or :-1: this comment to prefer manufacturerData and serviceData?

I can definitely see the value in being able to specify a mask in manufacturer data. If you are planning on supporting that in the future. Maybe it's better that we define the .manufacturerData filter to be a structure that contains (at this time only) manufacturer ID? If you don't opt for this, then it may be better to still keep the manufacturerId as the name of the field, then you can do a conversion or "deprecated detection" once you have a new version of the API.

nik-markovic:
>> 3) Shouldn't typical scan filters be applicable to this scan? For example, matching on a name.

> I think the name and namePrefix filters are the only ones I have in requestDevice() but not requestLEScan(). Android provides an exact name match, but not a partial name match, which limits the efficiency gains we can get by specifying a partial match. (Of course, iOS has much weaker filters.) My impression of requestDevice() uses is that most have needed to use partial name matches.

> So, I'm inclined not to include name filters in requestLEScan() until we find concrete use cases.

Isn't it implicit that requestLEScan filters are implemendted in software (browser) anyways? That way we are not limited to what the underlying bluetooth API implements. requestDevice is implemented in software, right? If we implement it in software, then we can have multiple sessions doing scans and filters of their own irrespective of what the adapter/native is filtering. Also, iOS and BluZ only supports limited filtering options, which typically include only service UUID and/or a couple of other minor things. How do we deal with that then? If this filter gets implemented in software, it seems to gravitate towards making filtering options and algorithm the same for both requestDevice and requestLEScan.

scanning.bs
jyasskin: Ok. I'll leave it out of the filters for now, and we can keep figuring out what to do on #253.

jyasskin:
> Isn't it implicit that requestLEScan filters are implemented in software (browser) anyways?

The browser has to fill in any filters that the platform doesn't provide, but the main reason to have a declarative system at all is so we can take advantage of OS and hardware support for the filters. Otherwise, we'd just let Javascript do the filtering.

On the other hand, having exactly the same filters for both systems would make them easier to understand, so I guess I'm not really opposed to doing that.

jyasskin:
The latest commits rename `manufacturerData` to `manufacturerId` and `serviceData` to `serviceDataUUID`, and add `name` and `namePrefix` filters. Any more comments?

Vudentz:
> Isn't it implicit that requestLEScan filters are implemented in software (browser) anyways?

Only doing filtering perhaps would be possible, but I guess if performance and security is a concern I would leave the filtering up to the underline stack so we don't spam the system with IPC messages containing advertisement reports that the browser may not be able to understand.

jyasskin:
The diversity of platforms means we have to expect to do some filtering in the browser. We should try to pass off as much as possible to the underlying platform though.

Is the security issue just that processes that can intercept IPC messages will find out the advertisements? Can't such privileged processes create their own advertisement requests anyway?

jyasskin:
I'm going to commit the current state. Please file issues when you notice them, and I'll fix them in smaller patches.

Vudentz:
> Is the security issue just that processes that can intercept IPC messages will find out the advertisements?

No, I meant that the browser itself is not a privileged application which shouldn't be given direct access for things that have shared control such as scanning.

This is actually acknowledge in the Android API:

https://developer.android.com/reference/android/bluetooth/le/BluetoothLeScanner.html
- Note: Most of the scan methods here require BLUETOOTH_ADMIN permission. (Btw, I wouldn't be surprised if this gets bumped to BLUETOOTH_PRIVILEGED if used with SCAN_MODE_LOW_LATENCY)

In Linux this would translate to have CAP_NET_ADMIN a.k.a root and then using the kernel Bluetooth Management socket to control the scanning parameters and results.

jyasskin:
Ok, yes, I expect the non-system-supported filters to have to run at a lower power use and/or stop while the browser's in the background. Chrome does take the BLUETOOTH_ADMIN permission in order to make `requestDevice()` work at all.

nikhilkashid:
Hi, I wanted to know if there is any method to get the UUID, Major, Minor and other services like Battery and signal strength of the signal received without requiring to connect with the device. We wanted to use beacons which would normally be in a unconnectable sleep mode. From what I understand requestLEScan would be a good method to use however in chrome console this method is not showing up and is undefined within navigator.bluethooth

g-ortuno:
@nikhilkashid Right, requestLEScan is the right method for this. Sadly it's not yet implemented in Chrome. https://github.com/WebBluetoothCG/web-bluetooth/blob/master/implementation-status.md#scanning-api

nikhilkashid:
Thank you @g-ortuno. We wanted to run a few experiments and view how requestLEScan would work. Is there a way we can try doing this?

beaufortfrancois:
@nikhilkashid You can't for now.

f138150:
@jyasskin @beaufortfrancois  HI, I've a question, devices without Peripheral mode also supports Web Bluetooth API ? 

Does Web Bluetooth API works with the BLE phones with central mode only(without peripheral mode). I want to send sensor data to web page using Web Bluetooth API. My phone has 4.2 BLE (Not peripheral mode) is not detecting by API. I tried on some other phone which supports peripheral mode, and it appears in the list.

I am using Ubuntu 16.04 LTE, Google chrome 55 (developer edition), enabled Bluetooth flag, Also follow guidelines [(https://acassis.wordpress.com/2016/06/28/how-to-get-chrome-web-bluetooth-working-on-linux/)]

g-ortuno:
@f138150 Peripheral mode is how devices (heart rate monitors, beacons, etc.) tell nearby devices of their presence. If a device doesn't support Peripheral mode then it can't communicate its presence so the API won't be able to find it.

That said, it's surprising a 4.2 BLE device doesn't support Peripheral mode. What device is it?

arashd:
@beaufortfrancois would you be able to provide any rough estimate of when we could expect watchAdvertisements/scanning to be implemented? 

Wouldn't this would be incredibly useful in combination with a Physical Web referral in a use-case where the user wants to easily distinguish between several devices (say parking meters) by distance (rssi) and pinpoint the one he's right next to. I'm new to the space; please correct me if I'm misunderstanding.

beaufortfrancois:
@arashd I'm eager as well to see this coming ;)
However I can't share any rough ETA at the moment.

arashd:
@beaufortfrancois I understand!

scheib:
@arashd, there's no active chromium work on scanning now. Still desired,
and there's still active Web Bluetooth work, but scanning will take a
while. https://www.chromium.org/teams/device-team/device-okrs were our
stated goals ~3 months ago. We're not hitting the Bluetooth targets due to
other priorities. We'll be publishing OKRs again in July.

arashd:
Thank you for that clarification @scheib. I'll keep an eye out for the new OKRs!

PR #249, opened 2016-07-15
Update the spec for changes in the Permissions spec.

jyasskin:
Preview at https://rawgit.com/jyasskin/web-bluetooth-1/update-permissions/index.html

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1035,19 +1035,19 @@ spec: permissions
     <li>
       <a>Get the <code>BluetoothDevice</code> representing</a> <var>device</var>
       inside the <a>context object</a> and <var>storage</var>,
-      and let <var>deviceObjPromise</var> be the result.
+      and let |deviceObj| be the result.
     </li>
     <li>
-      Wait for <var>deviceObjPromise</var> to settle.
+      If the previous step threw an exception,
+      <a>reject</a> |promise| with that exception and abort these steps.
     </li>
     <li>
-      If <var>deviceObjPromise</var> is <a>fulfilled</a>,
-      set <code><var>status</var>.{{BluetoothPermissionResult/devices}}</code>
-      to a new {{FrozenArray}} containing the value of <var>deviceObjPromise</var>
+      Set <code><var>status</var>.{{BluetoothPermissionResult/devices}}</code>
+      to a new {{FrozenArray}} containing the value of |deviceObj|
       as its only element.
     </li>
     <li>
-      <a>Resolve</a> <var>promise</var> with <var>deviceObjPromise</var>.
+      <a>Resolve</a> <var>promise</var> with `undefined`.
     </li>
   </ol>
 
@@ -1303,8 +1303,7 @@ spec: permissions
                 <a>Get the `BluetoothDevice` representing</a>
                 <code><var>allowedDevice</var>@{{[[device]]}}</code>
                 within <code><var>global</var>.navigator.bluetooth</code>,
-                wait for the result to settle,
-                and add its value to <var>matchingDevices</var>.
+                and add the result to <var>matchingDevices</var>.
               </li>
             </ol>
           </li>
--- a/index.bs
+++ b/index.bs
@@ -79,6 +79,7 @@ spec: BLUETOOTH-ASSIGNED
 
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: dfn
+        text: current realm; url: current-realm
         text: fulfilled; url: sec-promise-objects
         text: realm; url: sec-code-realms
     type: method
@@ -93,10 +94,6 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
         text: TypedArray; url: sec-typedarray-constructors
 
-spec: permissions; urlPrefix: https://w3c.github.io/permissions/
-    type: dfn
-        text: permission store
-
 spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/#
     type: dfn
         text: a copy of the bytes held; url: dfn-get-buffer-source-copy
@@ -105,8 +102,8 @@ spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/#
 <pre class="link-defaults">
 spec: html
     type: dfn
-        text: triggered by user activation
         text: browsing context
+        text: current settings object
         text: environment settings object
         text: event handler idl attribute
         text: global object
@@ -115,6 +112,7 @@ spec: html
         text: queue a task
         text: relevant settings object
         text: responsible event loop
+        text: triggered by user activation
 spec: permissions
     type: dfn
         text: create a permission storage entry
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -485,9 +483,9 @@ spec: permissions
         and a device can easily expose a custom GATT service to make this easier.
       </p>
       <p>
-        This specification
-        <a lt="add an allowed Bluetooth device">generates a new device ID</a>
-        for each origin for a given device,
+        This specification <a href="note-device-id-tracking">suggests</a> that
+        the UA use different device IDs for a single device
+        when its user doesn't intend script to learn that it's a single device,
```
g-ortuno: Hmmm are you missing a "the" before `script` or should it be plural i.e. `scripts`? You do this further down so I'm not sure if it's a typo or not.

index.bs
```diff
@@ -966,28 +959,33 @@ spec: permissions
...
+      <pre highlight="js">
+      {
+        name: "bluetooth",
+        filters: <var>uuidFilters</var>,
+        optionalServices: <var>optionalServiceUUIDs</var>,
+      }
+      </pre>
+    </li>
+    <li>
+      Let |state| be |descriptor|'s <a>permission state</a>.
```
g-ortuno: Do you think it's obvious that `|state|` will be `'denied'` for insecure origins? I wonder if it's worth calling it out.

index.bs
```diff
@@ -1002,7 +1000,9 @@ spec: permissions
     </li>
     <li id="requestDevice-prompt">
       Even if <var>scanResult</var> is empty,
-      display a prompt to the user requesting that the user select a device from it.
+      <a>prompt the user to choose</a> one of the devices in |scanResult|,
+      associated with |descriptor|,
+      and let |device| be the result.
       The UA SHOULD show the user the human-readable name of each device.
```
g-ortuno: optional nit: I think putting this in one of the green boxes could improve readability.

index.bs
```diff
@@ -1012,20 +1012,17 @@ spec: permissions
...
-      <a>resolve</a> <var>promise</var> with `undefined`,
-      and abort these steps.
-    </li>
-    <li>
-      <a lt="add an allowed bluetooth device"
-      >Add <var>device</var> to <var>storage</var></a>
-      with the union of <var>requiredServiceUUIDs</var>
-      and <var>optionalServiceUUIDs</var> as <var>allowed services</var>.
+    <li class="note">
+      Choosing a |device| probably indicates that
```
g-ortuno: Out of curiosity: How come you just say `the action probably indicates that 'something' should happen` rather than explicitly saying `The UA MAY do 'something'`?

index.bs
```diff
@@ -1012,20 +1012,17 @@ spec: permissions
...
+      the user intends that device to appear in
+      the {{BluetoothPermissionData/allowedDevices}} list
+      of {{"bluetooth"}}'s <a>extra permission data</a> for
+      at least the <a>current settings object</a>,
+      and for all the services in the union of <var>requiredServiceUUIDs</var>
+      and <var>optionalServiceUUIDs</var>
+      to appear in its {{AllowedBluetoothDevice/allowedServices}} list.
```
g-ortuno: I wonder if not explicitly specifying that the services need to be appended to the existing sequence will cause other implementations to miss that detail.

index.bs
```diff
@@ -1254,6 +1252,27 @@ spec: permissions
...
           that holds a <a>Bluetooth device</a>.
         </p>
+        <p>
+          Distinct elements of {{BluetoothPermissionData/allowedDevices}}
+          must have different {{AllowedBluetoothDevice/[[device]]}}s
+          and different {{AllowedBluetoothDevice/deviceId}}s.
+        </p>
+        <div class="note" id="note-device-id-tracking">
+          <p>
+            A {{AllowedBluetoothDevice/deviceId}} allows script to track that
```
g-ortuno: allows **a** script or allows script**s**?

index.bs
```diff
@@ -1268,24 +1287,33 @@ spec: permissions
       <dt>
         <a>permission query algorithm</a>
       </dt>
-      <dd>
+      <dd algorithm>
```
g-ortuno: nit: I seems a bit weird that the query algorithm comes before the request algorithm given that, the request example came before the query example and that you first have to request before you can query.

index.bs
```diff
@@ -1303,8 +1331,7 @@ spec: permissions
                 <a>Get the `BluetoothDevice` representing</a>
                 <code><var>allowedDevice</var>@{{[[device]]}}</code>
                 within <code><var>global</var>.navigator.bluetooth</code>,
-                wait for the result to settle,
-                and add its value to <var>matchingDevices</var>.
+                and add the result to <var>matchingDevices</var>.
```
g-ortuno: Do you think it's worth calling out that this algorithm ignores the `optionalServices` field and doesn't update allowedServices?

index.bs
```diff
@@ -1318,16 +1345,74 @@ spec: permissions
...
       <dt>
         <a>permission revocation algorithm</a>
       </dt>
-      <dd>
-        <p class="issue">
-          The permissions API's revocation support is still in flux,
-          so this isn't filled in yet.
+      <dd algorithm>
+        <p>
+          To <dfn>revoke Bluetooth access</dfn> to devices the user no longer intends to expose,
```
g-ortuno: So in the case of Chrome this algorithm would run after a user revokes permission for a device or for a service through site settings, right?

jyasskin: Right.

index.bs
jyasskin: It's not a typo, but I'll change it anyway to avoid stumbles. I'm having trouble explaining the grammar here, but I think this use of "script" is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun) kinda like "code", while "scripts" refers to multiple individual files with script in them.

index.bs
jyasskin: Done. https://w3c.github.io/permissions/#permission-state now requires this by default, but it's worth being explicit.

index.bs
jyasskin: Green boxes are non-normative, while I'd meant this to be normative. That said, "should" doesn't actually require UAs to do anything, so making it a note doesn't actually hurt anything concrete.

index.bs
jyasskin: https://w3c.github.io/permissions/#prompt-the-user-to-choose already allows the UA (that is, the "may") to change the extra permission data arbitrarily. This note is an attempt to get folks to tend to do the same thing unless they have a reason not to, instead of just having to guess.

index.bs
jyasskin: It's possible. I've added ", in addition to any services that were already there".

index.bs
jyasskin: Changed to "a script".

index.bs
jyasskin: Sure, swapped.

index.bs
jyasskin: My guess is that folks won't expect `query()` to update anything, but I can call out that `desc.optionalServices` doesn't affect the result.

jyasskin:
Ok, done with the updates in response to your comments. Thanks for the review!

index.bs
g-ortuno: Interesting. I've never heard `script` be used like that.

index.bs
g-ortuno: A separate paragraph would've also been fine. Before it was just one contiguous paragraph and it was somewhat hard to parse what the actual step was.

index.bs
g-ortuno: Ah I see.

g-ortuno:
lgtm!

PR #261, opened 2016-08-02
s/substeps/sub-steps/g

tkybpp:
I found two expressions, "substeps" and "sub-steps".
How about unify them? If there is no reason to distinct.
There are more "sub-steps" in the document, so I changed "substeps" to "sub-steps".

-

I've read CONTRIBUTING.md and I've already joined to [the Community Group](https://www.w3.org/community/web-bluetooth/) as "Takuya Beppu".


tkybpp:
Thanks😄

PR #264, opened 2016-08-03
Make startNotifications() return the characteristic, and …

### LINKED ISSUE #176: Set up characteristicvaluechanged event handlers if startNotifications fulfills its promise
It may be better to set up `characteristicvaluechanged` event handlers only if `startNotifications` fulfills its promise.

This means example at http://webbluetoothcg.github.io/web-bluetooth/#example-a130a116 should be

```js
function handleHeartRateMeasurementCharacteristic(characteristic) {
  return characteristic.startNotifications()
  .then(() => {
    characteristic.addEventListener('characteristicvaluechanged', onHeartRateChanged);
  });
}
```

instead of 

```js
function handleHeartRateMeasurementCharacteristic(characteristic) {
  characteristic.addEventListener('characteristicvaluechanged', onHeartRateChanged);
  return characteristic.startNotifications();
}
```

It would be great as well to return the characteristic in the fulfilled promise instead of `undefined`.

jyasskin:
make the example listen to the event after notifications start.

Fixes #176.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/listen-event-after-start-notifications/index.bs#example-heart-rate-monitor

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -228,8 +228,11 @@ spec: permissions
         }
 
         function handleHeartRateMeasurementCharacteristic(characteristic) {
-          characteristic.addEventListener('<a event>characteristicvaluechanged</a>', onHeartRateChanged);
-          return characteristic.<a for="BluetoothRemoteGATTCharacteristic">startNotifications()</a>;
+          return characteristic.<a for="BluetoothRemoteGATTCharacteristic">startNotifications()</a>
+          .then(_ => {
+            characteristic.addEventListener('<a event>characteristicvaluechanged</a>',
+                                            onHeartRateChanged);
+          });
         }
 
         function onHeartRateChanged(event) {
@@ -2948,7 +2951,7 @@ spec: permissions
           getDescriptors(optional BluetoothDescriptorUUID descriptor);
         Promise&lt;DataView> readValue();
         Promise&lt;void> writeValue(BufferSource value);
-        Promise&lt;void> startNotifications();
+        Promise&lt;BluetoothRemoteGATTCharacteristic> startNotifications();
         Promise&lt;void> stopNotifications();
       };
       BluetoothRemoteGATTCharacteristic implements EventTarget;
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
May we apply same treatment for `stopNotifications()`?

index.bs
```diff
@@ -228,8 +228,11 @@ spec: permissions
         }
 
         function handleHeartRateMeasurementCharacteristic(characteristic) {
-          characteristic.addEventListener('<a event>characteristicvaluechanged</a>', onHeartRateChanged);
-          return characteristic.<a for="BluetoothRemoteGATTCharacteristic">startNotifications()</a>;
+          return characteristic.<a for="BluetoothRemoteGATTCharacteristic">startNotifications()</a>
+          .then(_ => {
```
beaufortfrancois: We may want to show `characteristic` there.

```js
  .then(characteristic => {
```

beaufortfrancois:
For info, WIP at https://codereview.chromium.org/2216623002/

jyasskin:
> May we apply same treatment for `stopNotifications()`?

What's the use case? It's easy enough to do, but will anyone take advantage of it?

index.bs
jyasskin: For this code, I don't think I'd ever write out `characteristic` there, since it's already available from the function parameter. It makes the function a little harder to read, since you have to figure out which parameter is shadowing what, and then realize that they have the same value anyway.

g-ortuno:
lgtm. Someone could do:

```js
// ...
  return char.stopNotifications();
})
.then(char => char.removeEventListener(...));
```

jyasskin:
Yes, although if you're trying to stop listening, you may as well remove the listener first. Oh well, I'll include it.

index.bs
beaufortfrancois: Can at least show `char` then just so that developer sees that something is actually returned when promise resolves?

jyasskin: Oops, missed this. Sure.

PR #265, opened 2016-08-04
Make requestDevice()'s filters match requestLEScan()'s filters.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/match-requestDevice-filters-to-scanning/index.bs#device-discovery

### DIFF
--- a/implementation-status.md
+++ b/implementation-status.md
@@ -15,6 +15,7 @@ Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows | iOS
 ------------------------- | :-------: | :-----: | :-: | :---: | :-----: | :-:
 Discovery                 | ✓         | ✓       | ✓   | ✓     | ✓       | See notes
 └ Name or prefix          | ✓         | ✓       | ✓   | ✓     | ✓       |
+└ Manufacturer/Service data |         |         |     |       |         |
 Referring Device (Physical Web) |     |         |     |       |         |
 Chooser UI                | ✓         | ✓       | ✓   | ✓     | ✓       |
 GATT Server Connect       | ✓         | ✓       | ✓   | ✓     |         |
--- a/index.bs
+++ b/index.bs
@@ -587,8 +587,15 @@ spec: permissions
       </li>
     </ul>
     <p>
+      Note that this means that
+      if a device changes its behavior significantly when it connects,
+      for example by not advertising its identifying manufacturer data anymore
+      and instead having the client discover some identifying GATT services,
+      you may need to include filters both both behaviors.
+    </p>
+    <p>
       After the user selects a device to pair with this origin,
-      the origin is allowed to access to any service whose UUID was listed
+      the origin is allowed to access any service whose UUID was listed
       in the {{BluetoothRequestDeviceFilter/services}} list
       in any element of <code>options.filters</code>
       or in <code>options.<dfn dict-member for="RequestDeviceOptions">optionalServices</dfn></code>.
--- a/index.bs
+++ b/index.bs
@@ -525,6 +525,8 @@ spec: permissions
       sequence&lt;BluetoothServiceUUID> services;
       DOMString name;
       DOMString namePrefix;
+      unsigned short manufacturerId;
+      BluetoothServiceUUID serviceDataUUID;
     };
 
     dictionary RequestDeviceOptions {
@@ -566,10 +568,21 @@ spec: permissions
       </li>
       <li>
         have a name equal to <dfn dict-member for="BluetoothRequestDeviceFilter">name</dfn>
-        if that member is present, and
+        if that member is present,
       </li>
       <li>
         have a name starting with <dfn dict-member for="BluetoothRequestDeviceFilter">namePrefix</dfn>
+        if that member is present,
+      </li>
+      <li>
+        advertise <a>manufacturer specific data</a> with
+        a company identifier code equal to
+        <dfn dict-member for="BluetoothRequestDeviceFilter">manufacturerId</dfn>
+        if that member is present, and
+      </li>
+      <li>
+        advertise <a>service data</a> with
+        a UUID equal to <dfn dict-member for="BluetoothRequestDeviceFilter">serviceDataUUID</dfn>
         if that member is present.
       </li>
     </ul>
@@ -730,6 +743,53 @@ spec: permissions
     </table>
   </div>
 
+  <div class="example" id="example-filter-by-manufacturer-service-data">
+    <p>
+      Say the devices in the <a href="#example-filter-by-services">previous example</a>
+      also advertise manufacturer or service data as follows:
+    </p>
+    <table class="data">
+      <thead><th>Device</th><th>Manufacturer Data</th><th>Service Data</th></thead>
+      <tr><td>D1</td><td>17: 01 02 03</td><td></td></tr>
+      <tr><td>D2</td><td></td><td>A: 01 02 03</td></tr>
+    </table>
+
+    <p>
+      The following table shows which devices the user can select between
+      for several values of <var>filters</var> passed to
+      <code>navigator.bluetooth.requestDevice({filters: <var>filters</var>})</code>.
+    </p>
+
+    <table class="data">
+      <thead><th><var>filters</var></th><th>Devices</th></thead>
+      <tr>
+        <td>
+          <pre highlight="js">
+            [{manufacturerId: 17}]
+          </pre>
+        </td>
+        <td>D1</td>
+      </tr>
+      <tr>
+        <td>
+          <pre highlight="js">
+            [{serviceDataUUID: A}]
+          </pre>
+        </td>
+        <td>D2</td>
+      </tr>
+      <tr>
+        <td>
+          <pre highlight="js">
+            [{manufacturerId: 17},
+             {serviceDataUUID: A}]
+          </pre>
+        </td>
+        <td>D1, D2</td>
+      </tr>
+    </table>
+  </div>
+
   <div class="example" id="example-disallowed-filters">
     <p>
       Filters that either accept or reject all possible devices cause {{TypeError}}s.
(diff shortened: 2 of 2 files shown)


implementation-status.md
```diff
@@ -25,6 +25,7 @@ GATT Server Disconnect    | ✓         | ✓       | ✓   | ✓     |
 Get Characteristics List  | ✓         | ✓       | 53  | ✓     |         |
 Device Disconnected Event | ✓         | ✓       | ✓   | ✓     |         |
 Get Primary Services List | 53        | 53      | 53  | 53    |         |
+Manufacturer/Service data filters |   |         |     |       |         |
```
beaufortfrancois: Could we add it below `└ Name or prefix ` second row as `└ Manufacturer/Service data` to make it clear it is a filter?

beaufortfrancois:
LGTM but nit

implementation-status.md
jyasskin: Thanks, done.

g-ortuno:
qq: Are you thinking of changing the current filters to match the masking behavior on Android? The Eddystone EID seems like a good use case for it.

jyasskin:
Yes, but I'm not yet sure how to spell it in Javascript, or whether the length restriction in the Android API is exactly what we want.

g-ortuno:
The length? You mean the scanning duration restriction?

Also are you planning on including that change in this CL? Just wanted to know if I should review this now.

jyasskin:
Sorry, I meant that [`ScanFilter.Builder.setManufacturerData(id, data, mask)`](https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html#setManufacturerData(int, byte[], byte[])) requires that the advertised data be exactly as long as the array you provide, rather than letting it be a prefix or suffix. For example, you can't scan for Eddystone URL packets using the Android filter format without also getting all other Eddystone types.

That change won't be in this PR. I'll figure it out and change both `scanning.bs` and `index.bs` in some other PR.

jfarfel:
> Sorry, I meant that ScanFilter.Builder.setManufacturerData(id, data, mask)
<https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html#setManufacturerData(int,%20byte%5B%5D,%20byte%5B%5D)>
requires
that the advertised data be exactly as long as the array you provide,
rather than letting it be a prefix or suffix.
I think the data you provide can be a prefix of the full Mfg Data in the
advertisement.

Also, Eddystone data is in Service Data, not Mfg Data. So you'd want
setServiceData(uuid,
data, mask)
<https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder.html#setServiceData(android.os.ParcelUuid,
byte[], byte[])>. But the same thing still applies - you can give a prefix
(but not a suffix). See code:
http://cs/android/frameworks/base/core/java/android/bluetooth/le/ScanFilter.java?l=354&rcl=fcaee13dee3ef23e9855b77154de266770a49705

> For example, you can't scan for Eddystone URL packets using the Android
filter format without also getting all other Eddystone types.
Because it can be a prefix, I believe that you can indeed do this (since
the type is in a consistent spot in the service data).

jyasskin:
Oh, yep, that'll work. I still want to commit this PR making the two sets of filters the same, but then I'll think about how to spell the masked prefix filter in JS.

index.bs
```diff
@@ -524,6 +524,8 @@ spec: permissions
       sequence&lt;BluetoothServiceUUID> services;
       DOMString name;
       DOMString namePrefix;
+      unsigned short manufacturerId;
```
g-ortuno: Would it make sense to make these two arrays? How else could you express that you want a beacon that is advertising data for two services or two manufacturer's ids?

index.bs
```diff
@@ -565,10 +567,21 @@ spec: permissions
...
         have a name equal to <dfn dict-member for="BluetoothRequestDeviceFilter">name</dfn>
-        if that member is present, and
+        if that member is present,
       </li>
       <li>
         have a name starting with <dfn dict-member for="BluetoothRequestDeviceFilter">namePrefix</dfn>
+        if that member is present,
+      </li>
+      <li>
+        advertise <a>manufacturer specific data</a> with
```
g-ortuno: Hmm I wonder if this is going to make things confusing when multiple pages start to connect to the same device. For example the quad copter that only advertises manufacturer data. You would have to have a filter for the service uuid in case you are connected and one for the manufacturer data in case you are not connected. But maybe it's just something developers will have to deal with?

index.bs
```diff
@@ -565,10 +567,21 @@ spec: permissions
...
+        a company identifier code equal to
+        <dfn dict-member for="BluetoothRequestDeviceFilter">manufacturerId</dfn>
+        if that member is present, and
+      </li>
+      <li>
+        advertise <a>service data</a> with
+        a UUID equal to <dfn dict-member for="BluetoothRequestDeviceFilter">serviceDataUUID</dfn>
         if that member is present.
       </li>
     </ul>
```
g-ortuno: nit: The line after this says "...the origin is allowed to access to any service".

g-ortuno:
Overall it looks good. Just a couple of comments.

index.bs
beaufortfrancois: Bluetooth Core Spec 4.0 says respectively in Sections 10.18.10 and 10.18.11 that these values are unique.

Value | Description | Information
------------------------- | :-------: | :-----: 
0x16 | Service Data (2 or more octets) |The first 2 octets contain the 16 bit Service UUID followed by additional service data

Value | Description | Information
------------------------- | :-------: | :-----: 
0xFF | Manufacturer Specific Data (2 or more octets) | The first 2 octets contain the Company Identifier Code followed by additional manufacturer specific data

g-ortuno: What volume and part? I can't find any 10.18 sections?

index.bs
jyasskin: I think they'll just have to deal with identifying the device in two different ways if it behaves significantly differently. I've added a paragraph here mentioning that.

index.bs
jyasskin: CSSv6 A.1.4 and A.1.11 don't say that a packet "shall not contain more than one instance", so it's possible we'd see multiples. However, folks seem to be avoiding the 4-byte overhead of switching services or manufacturers by just tacking whatever data they need into a single field, so I suspect we'd see little use from an array of IDs. Also, is it likely that you'd need both IDs in order to identify the class of devices? Wouldn't they generally advertise one thing that's unique to the device and maybe another generic one?

index.bs
jyasskin: Fixed.

index.bs
```diff
@@ -566,16 +568,34 @@ spec: permissions
...
     <p>
+      Note that this means that
+      if a device changes its behavior significantly when it connects,
+      for example by not advertising its identifying manufacturer data anymore
+      and instead having the client discover some identifying GATT services,
+      you may need to include filters both both behaviors.
```
g-ortuno: s/both both/for both/ ?

jyasskin: I promise I can normally type.

g-ortuno:
lgtm

index.bs
beaufortfrancois: FWIW, I can't easily set multiple manufacturer specific data fields using mbed BLE code.

```cpp
const static uint8_t AdvData[] = {0x01,0x02,0x03,0x04,0x05};
const static uint8_t AdvData2[] = {0x06,0x07,0x08,0x09,0x10};
BLE::Instance().gap().accumulateAdvertisingPayload(
    GapAdvertisingData::MANUFACTURER_SPECIFIC_DATA, AdvData, sizeof(AdvData));
BLE::Instance().gap().accumulateAdvertisingPayload(
    GapAdvertisingData::MANUFACTURER_SPECIFIC_DATA, AdvData2, sizeof(AdvData2));

File: index.bs (hunk 2 above)
Comment (beaufortfrancois) [inline index.bs]: @bluetooth-mdw would you mind helping us there? Do you know if Bluetooth 5 will allow developers to express that they want a beacon that is advertising data for two services or two manufacturer's ids?

Comment (jyasskin) [inline index.bs]: FWIW, I believe BT5 changes the situation by allowing much larger advertising packets, which is likely to increase the frequency that we see multiple different manufacturers' or services' data. I'm now inclined to follow @g-ortuno's suggestion here.

Comment (beaufortfrancois) [inline index.bs]: @jyasskin @g-ortuno I've successfully broadcasted two manufacturer data IDs (one in advertisement packet, one in scan response) and from what I can see BlueZ handles it properly.

btmon logs:
```
> HCI Event: LE Meta Event (0x3e) plen 16
      LE Advertising Report (0x02)
        Num reports: 1
        Event type: Scan response - SCAN_RSP (0x04)
        Address type: Random (0x01)
        Address: DA:23:48:6B:B1:F5 (Static)
        Data length: 4
        Company: not assigned (1027)
          Data: 
        RSSI: -45 dBm (0xd3)
@ Device Found: DA:23:48:6B:B1:F5 (2) rssi -45 flags 0x0000
        02 01 06 04 09 46 52 00 03 ff 01 02 03 ff 03 04  .....FR.........
```

bluetoothctl logs:
```
[bluetooth]# info DA:23:48:6B:B1:F5 
Device DA:23:48:6B:B1:F5
        Name: FR
        Alias: FR
        Paired: no
        Trusted: no
        Blocked: no
        Connected: no
        LegacyPairing: no
        ManufacturerData Key: 0x0201
        ManufacturerData Key: 0x0403 
```

DBUS logs:
```
            dict entry(
               string "ManufacturerData"
               variant                   array [
                     dict entry(
                        uint16 513
                        variant                            array [
                           ]
                     )
                     dict entry(
                        uint16 1027
                        variant                            array [
                           ]
                     )
                  ]
            )
```

PR #266, opened 2016-08-04
Include connected devices in scan results.

### LINKED ISSUE #237: requestDevice() not considering already connected devices
A device could already be connected to the machine that the UA runs on, prior to the `requestDevice()` call. When a connection is established, devices often stop advertising, so won't get found by scanning and won't be "found" by `requestDevice()`.

A connection could have been established, prior to the `requestDevice()` call, for various reasons:
- if another site has already connected to it,
- the platform or another app running on it connected to it,
- the device connected to the machine that the UA runs on (in case the machine was acting as a peripheral at the same time).

Looking at the [steps that the UA has to take](https://webbluetoothcg.github.io/web-bluetooth/#ref-for-dom-bluetooth-requestdevice-3) for `requestDevice()`, it seems that already-connected, non-advertising devices are ignored and will not be shown in the list. I think the UA should also match already-connected devices, using the GATT cache and possibly also the last known advertisement data to match.

jyasskin:
Fixes #237.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/requestdevice-consider-connected/index.bs#scan-for-devices

@martijnthe, see if the wording looks right.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1124,7 +1124,11 @@ spec: permissions
       with a set of UUIDs that was a superset of the UUIDs for the current scan,
       then the UA MAY return the result of that scan and abort these steps.
     </li>
-    <li>Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially empty.</li>
+    <li>
+      Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s,
+      initially equal to the set of devices that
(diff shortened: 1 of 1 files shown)


g-ortuno:
lgtm

martijnthe:
SGTM, thanks!

PR #267, opened 2016-08-05
Allow requestDevice() to include non-discoverable devices.

### LINKED ISSUE #141: Discover "non-discoverable" beacons
There should be a way to discover beacons that have just the "BR/EDR Not Supported" flag set, not any of the "discoverable" flags. An option in `navigator.requestDevice` could look like this.

```js
var uriBeaconUUID = BluetoothUUID.getService(0xFED8);
  filters: [{ services: [uriBeaconUUID] }],
  mode: 'findEverythingWeCan' 
}).then(...);
```

Background:
_ChromeOS assumes the point of discovering devices is to connect to them, and a non-discoverable device is probably not connectable either, and because BLE provides the "General Discovery Procedure" to search for devices, and that says "If the Flags AD type (see [Core Specification Supplement], Part A, Section 1.3) is present and either the LE General Discoverable Mode flag is set to one or the LE Limited Discoverable Mode flag is set to one then the Host shall consider the device as a discovered device, otherwise the advertising data shall be ignored."_

jyasskin:
The CoreBluetooth API doesn't allow the UA to check the discoverable flags, so
we can't require it in the spec.

Fixes #141.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/discover-non-discoverable/index.bs#scan-for-devices

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1126,7 +1126,8 @@ spec: permissions
     </li>
     <li>Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially empty.</li>
     <li>
-      If the UA supports the LE transport, perform the <a>General Discovery Procedure</a>
+      If the UA supports the LE transport, perform the <a>General Discovery Procedure</a>,
+      except that the UA may include devices that have no <a>Discoverable Mode</a> flag set,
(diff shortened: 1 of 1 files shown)


g-ortuno:
lgtm

PR #268, opened 2016-08-05
Fix nit in example code

beaufortfrancois:
I believe `sessionStorage.lastDevice` is unique, therefore it makes more sense to check for `length == 1`. If not, it should be at least `length >= 1` not `length > 1`.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/WebBluetoothCG/web-bluetooth/raw/beaufortfrancois-patch-1/index.bs#example-permission-api-query

R=@jyasskin

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1273,7 +1273,7 @@ spec: permissions
           name: "bluetooth",
           deviceId: sessionStorage.lastDevice,
         }).then(result => {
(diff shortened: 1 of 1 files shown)


jyasskin:
I'd mildly prefer `>=`, but yeah, `>` is just wrong, and `==` is ok because only one device should have a given ID.

PR #269, opened 2016-08-08
Use . instead of @ to refer to internal slots.

jyasskin:
Matches tc39/ecma262#591.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/remove-%40/index.bs.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -82,6 +82,7 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: dfn
         text: current realm; url: current-realm
         text: fulfilled; url: sec-promise-objects
+        text: internal slot; url: sec-object-internal-methods-and-internal-slots
         text: realm; url: sec-code-realms
     type: method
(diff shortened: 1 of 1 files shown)


scheib:
@ -> . LGTM, but I still think that readers will be slightly confused when reading this spec. Perhaps the terms 'Internal Slot' in table headers can link to a simple explanation paragraph (similar to the one you're removing), or the spec sections that state e.g. "Instances of Bluetooth are created with the internal slots described..." could be slightly expanded to explain the internal slot concept.

jyasskin:
I've linked "internal slot" everywhere. Unfortunately, the ES standard is down right now, but I'll ping you when it's back up to check that the target is readable enough.

jyasskin:
https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots is back.

scheib:
;) The internal slots links are great, thanks.

PR #270, opened 2016-08-09
Use "attribute" instead of "entity" when appropriate.

jyasskin:
These come from before I realized that "attribute" was the generic term for a service, characteristic, or descriptor.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/rename-entity-to-attribute/index.bs#service-change-events


scheib:
LGTM

PR #271, opened 2016-08-10
Mark internal slots that the spec always initializes.

jyasskin:
Marking them instead of saying their initial value is `undefined`, makes
it easier to see if I accidentally introduce an uninitialized path.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/mark-always-initialized-fields/index.bs#dom-bluetoothdevice-context-slot

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1655,14 +1655,14 @@ spec: permissions
       </thead>
       <tr>
         <td><dfn>\[[context]]</dfn></td>
-        <td>`undefined`</td>
+        <td>&lt;always set in prose></td>
         <td>
           The {{Bluetooth}} object that returned this {{BluetoothDevice}}.
         </td>
       </tr>
       <tr>
         <td><dfn>\[[representedDevice]]</dfn></td>
-        <td>`undefined`</td>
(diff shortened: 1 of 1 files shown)


g-ortuno:
Does the "<always initialized>" mean we always expect them to be set when initializing the device object?

jyasskin:
Yep. Any ideas for how to say that more clearly?

g-ortuno:
Can't think of anything less verbose :disappointed: Is that what's usually used in specs?

jyasskin:
We're in uncharted territory here. https://streams.spec.whatwg.org/ is the only spec I know of that uses these tables, and it doesn't initialize any slots there. @domenic, any suggestions?

jyasskin:
I could imagine `<initialized in prose>`, for instance.

domenic:
Maybe just "set in prose" or "always set in prose"? I think any of these are reasonable.

jyasskin:
I picked "always set in prose".

g-ortuno:
lgtm

PR #272, opened 2016-08-10
Throw an InvalidStateError when an invalid attribute is used.

### LINKED ISSUE #130: Describe what to do if a method is called on a removed attribute
Services, Characteristics, and Descriptors can be removed from a device while an execution context has an object referring to them. If a method is then called on that object, we need to specify that an exception is thrown.

If the same attribute is later re-added to the device, we should probably re-vivify its object.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/invalidated-attributes/index.bs.

This fixes #130.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2537,11 +2537,6 @@ spec: permissions
         </p>
         <ol>
           <li>
-            If <code><var>deviceObj</var>.gatt.{{BluetoothRemoteGATTServer/connected}}</code>
-            is `false`,
-            <a>reject</a> <var>promise</var> with a {{NetworkError}} and abort these steps.
-          </li>
-          <li>
             <a>Populate the Bluetooth cache</a> with entries matching the description.
           </li>
           <li>
@@ -2589,6 +2584,26 @@ spec: permissions
           </li>
         </ol>
       </div>
+
+      <div algorithm>
+        <p>
+          <dfn>Represented</dfn>(|obj|: Device or GATT Attribute) returns,
+          depending on the type of |obj|:
+        </p>
+        <dl class="switch">
+          <dt>{{BluetoothDevice}}</dt>
+          <dd><code>|obj|.{{[[representedDevice]]}}</code></dd>
+
+          <dt>{{BluetoothRemoteGATTService}}</dt>
+          <dd><code>|obj|.{{[[representedService]]}}</code></dd>
+
+          <dt>{{BluetoothRemoteGATTCharacteristic}}</dt>
+          <dd><code>|obj|.{{[[representedCharacteristic]]}}</code></dd>
+
+          <dt>{{BluetoothRemoteGATTDescriptor}}</dt>
+          <dd><code>|obj|.{{[[representedDescriptor]]}}</code></dd>
+        </dl>
+      </div>
     </section>
 
     <section>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3862,6 +3992,48 @@ spec: permissions
...
+                from <code>|global|.navigator.bluetooth.{{[[attributeInstanceMap]]}}</code>
+                that represents an attribute that
+                is in |invalidatedAttributes|.
+              </li>
+            </ol>
+          </li>
+          <li>
             Let <var>changedServices</var> be a set of <a>Service</a>s, initially empty.
           </li>
           <li>
```
g-ortuno: In step 4 it says: 
> If attribute [...] appears in both, [...] remove it from both.

But in step 8 it says:
> If service [...] appears in both remove it from both and added to `changedServices`.

Maybe I'm missing something but wouldn't step 4 remove all services that step 8 would have added to `changedServices`?

index.bs
```diff
@@ -3862,6 +3992,48 @@ spec: permissions
             remove it from both.
```
g-ortuno: If someone tried to write to a characteristic between step 2 & 3 I'm guessing we would fail with some error like "Invalid Handle" or something like that. Does it make sense to specify what should happen in that case?

index.bs
jyasskin: Step 4 intends to remove attributes whose only changes are in their values. "with the same definition" may not be clear enough about this.

index.bs
jyasskin: Yeah, I think it's the same as if you start a write between when the characteristic is removed and when the device sends out the ServiceChanged indication. This is why I changed `Invalid Handle` down in https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/invalidated-attributes/index.bs#error-handling.

index.bs
g-ortuno: Ah right. Though, Step 2 says to use

>  Primary Service Discovery, Relationship Discovery, Characteristic Discovery, and Characteristic Descriptor Discovery procedures

None of those procedures reads the values of the Characteristics or Descriptors. If I understand correctly you only get the Value Handles.

jyasskin: If you have a service like:

Before:
*  Service A:
   * Characteristic B

After:
* Service A:
   * Characteristic C

Then step 4 doesn't remove Service A from *removedAttributes* or *addedAttributes* since it has a different definition (the list of included services and characteristics is part of a service definition per 3.G.3.1), step 8 does remove it, and I believe no other step would remove it.

I think Included Services force step 8 to include the "add it to *changedServices*." bit. For Characteristics, I think step 9 would cover it.

g-ortuno: Ah I see. Then yeah I think a link to what it means to have the same definition would make things clearer. But also not related to this patch so OK to do in follow up

g-ortuno:
lgtm

g-ortuno:
Did you mean to merge instead of closing?

jyasskin:
I did merge, as c8c9c525035a6905bf26cee3601e8db220a75a8a, 5fec7a4c98952f7070ecc8f7084f2bc942358d01, and 001c3a8750c0e9782b19cadcf44c44d073dab0d1, but I confused github by pushing in the wrong order. I'll also refine the ServiceChanged steps in another patch.

g-ortuno:
SG. Could you rebase the other pull request?

jyasskin:
Done.

Issue #226, opened 2016-03-31
Treat devices as if they weren't bonded to websites that have access to them.


jyasskin:
Per the thread at https://lists.w3.org/Archives/Public/public-web-bluetooth/2016Mar/0007.html, we'd like to have websites treat peripherals as if they're not bonded. That means that sites will need to re-discover services, characteristics, and descriptors after disconnecting, and they'll need to re-subscribe to notifications when they re-connect (#220), and might miss some that are fired between the connection and the re-subscription.

krantik:
Re-subscribing to notifications after disconnect is fine, similar to how the browser caches content should it cache the services/characteristics/descriptors(scd's), at minimum it should fetch these values ahead instead of getting them when asked which will improve the experience in terms of timing and then monitor for changes.

PR #273, opened 2016-08-12
Treat sites as not bonded to their devices.

jyasskin:
Preview: https://api.csswg.org/bikeshed/?url=https://github.com/jyasskin/web-bluetooth-1/raw/not-bonded/index.bs

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2450,6 +2450,19 @@ spec: html
     </div>
 
     <section>
+      <h4 id="persistence">Persistence across connections</h4>
+
+      The Bluetooth <a>Attribute Caching</a> system allows <a>bonded</a> clients
+      to save references to attributes from one connection to the next.
+      Web Bluetooth treats websites as <em>not</em> <a>bonded</a>
+      to devices they have permission to access:
+      {{BluetoothRemoteGATTService}}, {{BluetoothRemoteGATTCharacteristic}},
+      and {{BluetoothRemoteGATTDescriptor}} objects
+      become invalid on <a href="#disconnection-events">disconnection</a>,
+      and the site must retrieved them again when it re-connects.
+    </section>
+
+    <section>
       <h4 id="bluetooth-cache" dfn-type="dfn" data-lt="Bluetooth cache">The Bluetooth cache</h4>
 
       <p>
@@ -2640,7 +2653,8 @@ spec: html
             return <a>a promise rejected with</a> an {{InvalidStateError}} and
             abort these steps.
 
-            Note: This happens when a Service or Characteristic is removed from the device,
+            Note: This happens when a Service or Characteristic is removed from the device
+            or invalidated by a disconnection,
             and then its object is used again.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
Does invalidating attributes when a device disconnects means attribute becomes `null`? If so, it makes it harder to use a pattern like the one below:

```js
var characteristic

function doSomething() {
  if (!characteristic) {
    return navigator.bluetooth.requestDevice({
        filters: [{
          services: ['battery_service']
        }]
      })
      .then(device => device.gatt.connect())
      .then(server => server.getPrimaryService('battery_service'))
      .then(service => service.getCharacteristic('battery_level'))
      .then(char => {
        characteristic = char;
      });
  }
  return Promise.resolve()
    .then(_ => {
      if (!characteristic.service.device.gatt.connected) {
        // Device is not connected anymore, let's reconnect.
        return characteristic.service.device.gatt.connect();
      }
    })
    .then(_ => characteristic.readValue());
}

**Comment by jyasskin** [comment]
Yes, folks will need to write a custom `connect()` function that discovers the attributes they need on each connection. Something like:

```js
let device;
let characteristic;

function reconnect() {
  if (device.gatt.connected) return Promise.resolve();
  return device.gatt.connect()
      .then(server => server.getPrimaryService('battery_service'))
      .then(service => service.getCharacteristic('battery_level'))
      .then(char => { characteristic = char; });
}

function doSomething() {
  let dp = Promise.resolve();
  if (!device) {
    dp = navigator.bluetooth.requestDevice({
        filters: [{
          services: ['battery_service']
        }]
      })
      .then(d => device = d.gatt.connect());
  }
  return dp.then(() => reconnect())
    .then(_ => characteristic.readValue());
}
```

g-ortuno:
Wrong preview link :smile:

beaufortfrancois:
@jyasskin `d.gatt.connect()` returns a GATT Server instance. Did you mean?

```diff
-      .then(d => device = d.gatt.connect());
+      .then(d => { device = d; return device.gatt.connect() });
````

**Comment by jyasskin** [comment]
@g-ortuno Oops, fixed.
@beaufortfrancois Oops, I meant `device=d;`. Then `reconnect()` handles connecting.

File: index.bs [hunk 1]
```diff
@@ -4004,13 +4007,46 @@ spec: permissions
...
+        </ol>
+      </div>
+
+      <div algorithm>
+        <p>
+          To <dfn>clean up the disconnected device</dfn> |deviceObj|, the UA must:
+        </p>
+        <ol>
+          <li>
             Set <code><var>deviceObj</var>.gatt.{{BluetoothRemoteGATTServer/connected}}</code>
```
g-ortuno: You set this twice when this is called from `disconnect()`. Maybe remove the one in `disconnect()`?

g-ortuno:
lgtm bar a nit.

g-ortuno:
I wonder if you still need the ActiveAlgorithms internal slot. Now that we invalidate attributes on disconnection maybe you can just check if `characteristic.[[representedCharacteristic]]` is null.

Vudentz:
I guess the bonded terminology here is not used in the same way as in Bluetooth, so it means on disconnected all cached attributes are flushed. This would work for most cases except if the profile requires notification/indications to be active the entire time, in that case flushing the cache means the application won't be able to receive any value changed until it reconnects, discover the attributes and subscribe to value changed which may take some time causing the application to miss values.

Also, how about if there are multiple sites connected to the same device? Doesn't that still cause the same situation where cached attributes are available?

jyasskin:
@Vudentz In Bluetooth, if a device isn't bonded, when it disconnects the attribute cache also needs to be flushed, according to 3.G.2.5.2:

> For clients that have a trusted relationship (i.e. bond) with the server, the attribute cache is valid across connections. For clients with a trusted relationship and not in a connection when a service change occurs, the server shall send an indication when the client reconnects to the server. For clients that do not have a trusted relationship with the server, the attribute cache is valid only during the connection. 

So I think the terminology here is exactly in line with Bluetooth.

Yes, I called out the risk of missing notifications in https://github.com/WebBluetoothCG/web-bluetooth/issues/226.

If sites X, Y, and Z are connected, and X disconnects, X's attributes will get invalidated even though the platform is still connected and could avoid invalidating the attributes. That's the right thing to do so that developers don't have to/aren't tempted to write two different code paths for the two different cases.

index.bs
jyasskin: It's removed: https://github.com/WebBluetoothCG/web-bluetooth/pull/273/files#diff-ec9cfa5f3f35ec1f84feb2e59686c34dL2853. Were you looking at `If <code>this.{{connected}}</code> is <code>false</code>`?

g-ortuno: Oops maybe I was looking at the wrong version of the spec.

jyasskin:
Removing `[[activeAlgorithms]]` is a really good idea that almost works, but it turns out that we need something like it to track disconnections during `getPrimaryService()`, and since the GATTServer isn't replaced across disconnection, there's not an obvious place to put that marker. :disappointed: I'll leave it here for this patch, but maybe we'll find a way in the future to remove or simplify it.

PR #278, opened 2016-08-29
Add a way to include all devices in the chooser.

### LINKED ISSUE #234: Allow requestDevice for all devices

jyasskin:
`requestDevice({acceptAllDevices: true})` matches Scanning's
`requestLEScan({acceptAllAdvertisements: true})`.

Fixes #234.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#device-discovery

@domenic, I'm not happy with how I expressed that exactly one of two fields in a dictionary must be present, especially since one of the fields is actually a boolean that's always present but I actually mean "true" for it. Do you have any suggestions? Also, "flag" arguments to algorithms: should I be using a boolean instead?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1025,7 +1025,7 @@ spec: html
         <a>Request Bluetooth devices</a>,
         passing <code>|options|.{{filters}}</code>
         if <code>|options|.{{acceptAllDevices}}</code> is `false`
-        or the <i>acceptAllDevices</i> flag otherwise,
+        or `null` otherwise,
         and passing <code>|options|.{{optionalServices}}</code>,
         and let |devices| be the result.
       </li>
@@ -1046,8 +1046,8 @@ spec: html
   <div algorithm>
     <p>
       To <dfn>request Bluetooth devices</dfn>,
-      given either a sequence of {{BluetoothRequestDeviceFilter}}s, |filters|
-      or an |acceptAllDevices| flag,
+      given a sequence of {{BluetoothRequestDeviceFilter}}s, |filters|,
+      which can be `null` to represent that all devices can match,
       and a sequence of {{BluetoothServiceUUID}}s, |optionalServices|,
       the UA MUST run the following steps:
     </p>
@@ -1069,8 +1069,7 @@ spec: html
         do the following sub-steps:
         <ol>
           <li>
-            If |acceptAllDevices| isn't set
-            and <code>|filters|.length === 0</code>,
+            If <code>|filters| !== null && |filters|.length === 0</code>,
             throw a {{TypeError}}
             and abort these steps.
           </li>
@@ -1079,11 +1078,11 @@ spec: html
             <var>requiredServiceUUIDs</var> be a new {{Set}}.
           </li>
           <li>
-            If |acceptAllDevices| is set,
+            If |filters| is `null`,
             then set |requiredServiceUUIDs| to the set of all UUIDs.
           </li>
           <li>
-            If |acceptAllDevices| isn't set,
+            If |filters| isn't `null`,
             then for each |filter| in |filters|,
             do the following steps:
             <ol>
@@ -1213,18 +1212,13 @@ spec: html
         </ol>
       </li>
       <li>
-        Let |acceptAllDevicesBool| be `true`
-        if the <var>acceptAllDevices</var> flag is set</i>,
-        or `false` otherwise.
-      </li>
-      <li>
         Let |descriptor| be
         <pre highlight="js">
         {
           name: "bluetooth",
           filters: <var>uuidFilters</var>,
           optionalServices: <var>optionalServiceUUIDs</var>,
-          acceptAllDevices: <var>acceptAllDevicesBool</var>,
+          acceptAllDevices: <var>filters</var> !== null,
         }
         </pre>
       </li>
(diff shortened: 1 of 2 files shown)


domenic:
Interesting, we were just discussing MutationObservers, which have a similar dictionary. Per Web IDL, the `options` argument must be optional, even though the prose will throw if nothing is passed. (We make cases like this, and like MutationObservers, add a redundant `optional`, so we can enforce the more generally-applicable rule that trailing dictionaries must be optional, since most of them don't have these kind of "at least one must be present" rules.)

This algorithm never seems to go "in parallel" so it's unclear why a promise is being used here to wrap synchronous exceptions/return values. I guess that's a separate issue; "request Bluetooth devices" should go "in parallel" instead of synchronously performing all its work.

Normally I'd go with flags for consistency but in this case I see you have to convert it to a boolean later passing a boolean directly makes sense. (I wish we used booleans everywhere but we don't.)

One way to handle this might be for the internal algorithm to take a sequence of filters or null and use null to represent all devices. You could state it explicitly as "To request Bluetooth devices, given either a sequence of BluetoothRequestDeviceFilters or null _filters_ (where null represents accepting all devices) and and a sequence of BluetoothServiceUUIDs _optionalServices_, the UA MUST run the following steps:". This could clean up the internals a bit.

I think the actual requestDevice algorithm is fine. I even kind of like the way that the boolean defaulting to false means you can assume it's always present, which I guess you are unsure about. But it seems nice to me.

jyasskin:
Thanks @domenic! I've fixed the 'optional' problem; that was just a mistake on my part.

["Request Bluetooth devices"](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#request-bluetooth-devices) does a long task in step 7 and waits for user input in step 9, so it can only be used "in parallel". It was easier to write this way, since both callers use it in parallel (see [requestDevice(), which uses a shorthand you don't really like](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#dom-bluetooth-requestdevice) and [permissions.request(), which uses the archaic "continue the following steps asynchronously."](https://w3c.github.io/permissions/#dom-permissions-request)). I failed to mark that in note like I've done [elsewhere](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#populate-the-bluetooth-cache).

I was avoiding "no filters" as the code for "accept all devices" because it leads to the odd situation where 0 filters returns more devices than 1 filter, which returns fewer devices than 2 filters. However, since the ["request Bluetooth devices"](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#request-bluetooth-devices) algorithm is internal, maybe that sort of ergonomics matters less. This isn't done yet, but I'll start on it.

Thanks for taking a look!

index.bs
```diff
@@ -582,6 +583,15 @@ spec: html
       the website may need to include filters for both behaviors.
     </p>
     <p>
+      In rare cases,
+      a device may not advertise enough distinguishing information
+      to let a site filter out uninteresting devices.
+      In those cases, a site can set
+      <dfn dict-member for="RequestDeviceOptions">acceptAllDevices</dfn> to `true`
+      and omit all {{RequestDeviceOptions/filters}}.
+      This puts the burden of selecting the right device entirely on the site's users.
```
beaufortfrancois: Do you mind specifying (again) that developer will have to use `optionalServices` to access GATT Services if they choose to set `acceptAllDevices` to `true` as it overrides any `filters`?

jyasskin: How's https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/open-filter/index.bs#dom-requestdeviceoptions-acceptalldevices now?

beaufortfrancois: Perfect!

PR #279, opened 2016-08-30
Clear the active notification context set on disconnection.

### LINKED ISSUE #220: startNotifications should always write descriptor if device is not bonded
If devices are not bonded, notifications won't be remembered after a disconnection.

Therefore, instead of giving the responsibility the check whether devices are bonded or not to the UA, I suggest `startNotifications()` should always write descriptor no matter what.

Same applies to `stopNotifications()` of course.

WDYT?

Background: https://bugs.chromium.org/p/chromium/issues/detail?id=589796#c9

jyasskin:
Fixes #220.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/clear-notifications-on-disconnection/index.bs#clean-up-the-disconnected-device.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3441,6 +3441,12 @@ spec: html
         The set for a given characteristic holds
         the {{Navigator/bluetooth|navigator.bluetooth}} objects for
         each <a>Realm</a> that has registered for notifications.
+        All notifications become inactive when a device is disconnected.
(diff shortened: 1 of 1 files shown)


scheib:
LGTM

PR #280, opened 2016-08-31
Make disconnect() abort connect().

### LINKED ISSUE #152: Should connectGATT() time out or wait until a device is available?
And if it waits, should that wait be cancellable?

On at least [Android](http://developer.android.com/reference/android/bluetooth/BluetoothGatt.html#connect()) and [iOS](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/index.html#//apple_ref/occ/instm/CBCentralManager/connectPeripheral:options:), if you call "connect" on a device, and the device isn't nearby, the call just waits until the device shows up. Bluetooth specifies T<sub>GAP</sub>(adv_slow_interval) <= 1.2s as the longest the central's supposed to have to wait to connect to a present device, so either we could time out the connects, or users could. If we make users do it, it naturally solves #31, but we probably want a way to cancel the attempt to connect. Right now, unlike in the Android and iOS APIs, we only expose [`disconnect()` on the `BluetoothGATTRemoteServer`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothgattremoteserver-disconnect), which is only available once [`connectGATT()`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-connectgatt)'s promise fulfills.

Right now, I'm inclined to say that `connectGATT()` should hang until a device shows up, mention in a Note that 1.2s makes a good timeout, and give some way to cancel the connection, maybe by making `connectGATT()` fill in `BluetoothDevice.gattServer` synchronously, and having folks call `disconnect()` on that.

jyasskin:
Also add a note that connect() could wait forever if an appropriate
advertisement isn't received.

Fixes #152.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/hanging-connect/index.bs#dom-bluetoothremotegattserver-connect.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2896,35 +2896,74 @@ spec: html
     <div algorithm>
       <p>
         The <code><dfn method for="BluetoothRemoteGATTServer">connect()</dfn></code> method, when invoked,
-        MUST return <a>a new promise</a> <var>promise</var> and
-        run the following steps <a>in parallel</a>:
+        MUST perform the following steps:
       </p>
       <ol>
+        <li>Let |promise| be <a>a new promise</a>.</li>
         <li>
           If <code>this.device.{{[[representedDevice]]}}</code> is `null`,
-          <a>reject</a> <var>promise</var> with a {{NetworkError}} and abort these steps.
+          <a>queue a task</a> to <a>reject</a> |promise| with a {{NetworkError}},
+          return |promise|,
+          and abort these steps.
         </li>
         <li>
-          If <code>this.device.{{[[representedDevice]]}}</code> has no <a>ATT Bearer</a>,
-          attempt to create one using the procedures described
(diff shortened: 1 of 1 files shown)


scheib:
I see that this is also specifying that disconnect will not sever other
connections on the system. E.g. if 2 Android apps are using a connection to
a peripheral and one of them is a Web Bluetooth browser, then calling
disconnect there will not disconnect the central from the peripheral.
Sounds good to me, but worth perhaps calling that out in the change
description as well.

Overall LGTM

jyasskin:
I've updated the change description to mention the clarification on the `disconnect()` behavior.

PR #281, opened 2016-09-01
BluetoothDevice.gatt is not optional

beaufortfrancois:
See https://chromium.googlesource.com/chromium/src/+/master/third_party/WebKit/Source/modules/bluetooth/BluetoothDevice.idl

Live preview at https://api.csswg.org/bikeshed/?url=https://github.com/WebBluetoothCG/web-bluetooth/raw/beaufortfrancois-patch-1/index.bs#bluetoothdevice

R: @jyasskin

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1697,7 +1697,7 @@ spec: html
       interface BluetoothDevice {
         readonly attribute DOMString id;
         readonly attribute DOMString? name;
-        readonly attribute BluetoothRemoteGATTServer? gatt;
+        readonly attribute BluetoothRemoteGATTServer gatt;
         readonly attribute FrozenArray&lt;UUID> uuids;
 
         Promise&lt;void> watchAdvertisements();


jyasskin:
`BluetoothDevice.gatt` is always filled in by this spec (actually, maybe not: should I fail `requestDevice()` if it doesn't request any services?), but the [Scanning spec](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#advertising-events) leaves it `null` by setting no allowed services. https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-gatt

beaufortfrancois:
I've just tried requesting a device with just a device name filter at https://googlechrome.github.io/samples/web-bluetooth/device-info.html and `device.gatt` is NOT null. Should we update our implementation then?

jyasskin:
Yep, we should update our implementation. I could also arrange for scanning to mark that it doesn't grant permission to connect in some other way, if that's better. The important bit is that scanning leave `.gatt` null until the site upgrades in some way.

g-ortuno:
I don't think it should be null. If you are a developer and you forget to add optionalServices then `.gatt` will be null and there will be no indication as to why it's null. If we allow users to connect then at least they will get an error message when they try to retrieve the services.

g-ortuno:
^ for the general case. I think it makes sense for the Scanning API to leave it null'ed

jyasskin:
For `requestDevice()`, I'm tempted to specify that if you don't have any allowed services, then you get an error, instead of silently getting a useless device. Would that make sense?

g-ortuno:
That would break our device info demo :(

jyasskin:
Arguably, that one should fetch data from https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.service.device_information.xml in addition to the global fields. But sure, I can use a dedicated flag to say whether a site has permission to connect, instead of overloading `[[allowedServices]]`.

beaufortfrancois:
> For requestDevice(), I'm tempted to specify that if you don't have any allowed services, then you get an error, instead of silently getting a useless device. Would that make sense?

Nope. Device won't get useless.
How about watching advertisements simply from a device?

```js
navigator.bluetooth.requestDevice({filters: [{name: "Jeffrey's Robot"}]})
  .then(device => {
    console.log('> Name:             ' + device.name);
    console.log('> Id:               ' + device.id);
    console.log('> Allowed Services: ' + device.uuids.join('\n' + ' '.repeat(20)));
    console.log('> Connected:        ' + device.gatt.connected);
    // Start watching advertisements...
    device.watchAdvertisements();
  })

beaufortfrancois:
Since `gatt` can be `null` now, that means developers have to check its value before using its methods right? If so, I'll update our [device info sample](https://googlechrome.github.io/samples/web-bluetooth/device-info.html) then.

```js
navigator.bluetooth.requestDevice({filters: [{name: "Jeffrey's Robot"}]})
  .then(device => {
    console.log('> Name:             ' + device.name);
    console.log('> Id:               ' + device.id);
    if (device.gatt) {
      console.log('> Connected:        ' + device.gatt.connected);
    }
  })

jyasskin:
@beaufortfrancois If you've gotten a device via `requestDevice()`, you're guaranteed `.gatt` isn't `null`, at least until the user revokes your access. Some of this might change once we have the upgrade from scanning to communication, but that's a while from now.

PR #287, opened 2016-09-03
Expose whether Bluetooth is available.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/expose-availability/index.bs#availability.

Thanks to @domenic for simplifying the overall design.

@npdoty can you see if you agree with the [privacy section](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/expose-availability/index.bs#availability-fingerprint) or if there's something different I should say?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -533,6 +533,10 @@ spec: html
 
     interface Bluetooth {
       [SecureContext]
+      Promise&lt;boolean> getAvailability();
+      [SecureContext]
+      attribute EventHandler onavailabilitychanged;
+      [SecureContext]
       readonly attribute BluetoothDevice? referringDevice;
       [SecureContext]
       Promise&lt;BluetoothDevice> requestDevice(optional RequestDeviceOptions options);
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1684,6 +1717,100 @@ spec: html
...
+      Note: the promise is resolved <a>in parallel</a>
+      to let the UA call out to other systems
+      to determine whether Bluetooth is available.
+    </div>
+
+    <div algorithm="availabilitychanged" id="availability-changed-algorithm">
+      <p>
+        If the UA becomes able or unable to use Bluetooth,
+        for example because a radio was physically attached or detached,
+        the UA must run the following steps:
```
scheib: Must? What about for UAs that choose to not disclose the availability?

scheib:
Looks good to me with the exception noted regarding UAs that choose to not
disclose the availability: the wording for availabilitychanged event should
be clear that UAs may choose to not implement this.

index.bs
```diff
@@ -1684,6 +1717,100 @@ spec: html
...
+      but websites may be able to handle it more gracefully.
+    </p>
+
+    <div class="example" id="example-getavailability">
+      <p>
+        To show Bluetooth UI only to users who have a Bluetooth adapter:
+      </p>
+      <pre highlight="js">
+      let bluetoothUI = document.querySelector('#bluetoothUI');
+      navigator.bluetooth.<a idl>getAvailability()</a>.then(available => {
```
beaufortfrancois: I think `isAvailable` name would make it clearer this is a boolean value instead of just `available`.

index.bs
jyasskin: Good point.

jyasskin: Done, by talking about the user configuring the UA to return a particular value.

index.bs

beaufortfrancois:
👍

jyasskin:
@msramek from Chrome's privacy team thinks this is fine, so I'm going to go ahead and submit it.

PR #289, opened 2016-09-08
Make the tests spec auto-build, and fix the link to it.

jyasskin:
I'm not sure if I should fix the link or remove it. Thoughts?


g-ortuno:
lgtm. I think it's OK to leave it. We will soon start working on the newer version of the API.

PR #292, opened 2016-09-09
Remove the BluetoothDevice.uuids field.

### LINKED ISSUE #290: How to handle changing advertised UUIDs
Step 8 of [get the BluetoothDevice representing](https://webbluetoothcg.github.io/web-bluetooth/#get-the-bluetoothdevice-representing)  states:
> 8. Add device’s advertised Service UUIDs to result.[[unfilteredUuids]].

But it isn't clear what advertised Service UUIDs means when a device is alternating UUIDs. Is it only the last Advertised UUIDs? All the UUIDs we've seen during that particular scan or all the UUIDs we've seen so far across scans?

Given that only BlueZ caches the advertised UUIDs I think this should be specified.

jyasskin:
It was hard to define the value of this field when a device's
advertisements may list varying sets of services and when the primary
services in the GATT server may be discovered gradually. Developers can
replace their uses of this field with uses of getPrimaryServices().

Fixes #290 

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/remove-uuids/index.bs. Please double-check that there's nothing else I can remove now that this field is gone.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1923,7 +1923,6 @@ spec: html
         readonly attribute DOMString id;
         readonly attribute DOMString? name;
         readonly attribute BluetoothRemoteGATTServer? gatt;
-        readonly attribute FrozenArray&lt;UUID> uuids;
 
         Promise&lt;void> watchAdvertisements();
         void unwatchAdvertisements();
@@ -1951,12 +1950,6 @@ spec: html
       <p>
         {{BluetoothDevice/gatt}} provides a way to interact with this device's GATT server
         if the site has permission to use any of its services.
-      </p>
-
-      <p>
-        {{BluetoothDevice/uuids}} lists
-        the UUIDs of GATT services known to be on the device,
-        that the current origin is allowed to access.
       </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


Vudentz:
I guess the main reason to aggregate the UUIDs is that GATT and GAP may have a completely different set of UUIDs.

jyasskin:
Can you point me to the description of GAP UUIDs in the Bluetooth spec? I know about SDP UUIDs, but we'll add those if and when we add Bluetooth Classic support, and they'll do fine in a list that's distinct from the GATT list.

Vudentz:
UUIDs are not specific to GATT/GAP, what I meant is that advertisement UUIDs may not be the same as you can find in GATT, in fact I believe this is already the case for 16-bit UUIDs for members which may appear in the advertisement but not in any service or in case the UUID is for a service not on GATT but on a custom L2CAP channel.

thegecko:
We too use `BluetoothDevice.uuids` to store the services advertised by the device as we always assumed this could be a subset of the services discoverable using `BluetoothRemoteGATTServer.getPrimaryServices()`. If they are guaranteed to be the same, I see the point of this proposal.

jyasskin:
To get the advertised UUIDs, sites can call [<code>*device*.watchAdvertisements()</code>](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-watchadvertisements), but that does miss the data that was sent before the device was paired. Doing that lets the site ask for a precise behavior when the device is sending advertisements with more than one value. For example, devices can alternate advertisements that contain an Eddystone URL with advertisements that contain a list of Service UUIDs.

If we want to expose the content of advertisements from before the device was paired, how should we expose it? Should we ask the UA to union the UUIDs from some set of previous advertisements, or only include the most recent one? Should `device.uuids` also include discovered services or just advertised ones? I'm inclined to remove it so that we can take our time answering those questions, but if there's a good argument for a particular answer, I'm happy to take that instead.

jyasskin:
Hearing no other comments, I'd like to get a review for the details of the change. We can add `uuids` back if we come up with some more precise semantics for it.

beaufortfrancois:
See https://github.com/GoogleChrome/samples/pull/422

PR #293, opened 2016-09-09
Specify that UAs should negotiate the MTU.

### LINKED ISSUE #284: Add Web Bluetooth API support to change MTU
I've noticed Chrome OS and Android have different maximum transmission unit (MTU) when investigating https://bugs.chromium.org/p/chromium/issues/detail?id=643591 bug and I was wondering if we could add support for requesting an MTU size used for a given connection in the Web Bluetooth API.

Android already has this API which is really useful in my own opinion: https://developer.android.com/reference/android/bluetooth/BluetoothGatt.html#requestMtu(int)

For info, 23 bytes seems to be the default on Android while I can reach 512 bytes on Chrome OS.

WDYT?

jyasskin:
Fixes #284 

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/max-mtu/index.bs#dom-bluetoothremotegattserver-connect.

@beaufortfrancois @Vudentz

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2922,6 +2922,10 @@ spec: html
                   If this attempt failed for another reason,
                   <a>reject</a> <var>promise</var> with a {{NetworkError}} and abort these steps.
                 </li>
+                <li>
+                  Use the <a>Exchange MTU</a> procedure to negotiate the largest supported MTU.
+                  Ignore any errors from this step.
+                </li>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
LGTM

PR #297, opened 2016-09-23
Accept filters on manufacturer and service data.

### LINKED ISSUE #294: Support for requesting devices with specific service data in adv packets
Is there any plan to add filtering by specific service data? I mean something like bit mask filter for requesting devices with specific pattern inside the service data.

jyasskin:
Previews at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/masked-data-prefixes/index.bs and https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/masked-data-prefixes/scanning.bs.

Fixes #294. @irgu

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -81,6 +81,7 @@ spec: BLUETOOTH-ASSIGNED
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
     type: abstract-op
         text: CanonicalNumericIndexString; url: sec-canonicalnumericindexstring
+        text: CreateDataProperty; url: sec-createdataproperty
         text: IsInteger; url: sec-isinteger
     type: dfn
         text: current realm; url: current-realm
@@ -1486,10 +1487,14 @@ spec: html
             If this conversion throws an exception, propagate it and abort these steps.
           </li>
           <li>
-            Set <code>|canonicalizedFilter|.manufacturerData[|manufacturerId|]</code> to be
+            Let |canonicalizedDataFilter| be
             the result of <a for="BluetoothDataFilterInit">canonicalizing</a> |dataFilter|,
             <a>converted to an ECMAScript value</a>.
             If this throws an exception, propagate that exception and abort these steps.
+          </li>
+          <li>
+            Call <a abstract-op>CreateDataProperty</a>(|canonicalizedFilter|.manufacturerData,
+            |key|, |canonicalizedDataFilter|).
           </li>
         </ol>
       </li>
@@ -1529,10 +1534,14 @@ spec: html
             If this conversion throws an exception, propagate it and abort these steps.
           </li>
           <li>
-            Set <code>|canonicalizedFilter|.serviceData[|service|]</code> to be
+            Let |canonicalizedDataFilter| be
             the result of <a for="BluetoothDataFilterInit">canonicalizing</a> |dataFilter|,
             <a>converted to an ECMAScript value</a>.
             If this throws an exception, propagate that exception and abort these steps.
+          </li>
+          <li>
+            Call <a abstract-op>CreateDataProperty</a>(|canonicalizedFilter|.serviceData,
+            |service|, |canonicalizedDataFilter|).
           </li>
         </ol>
       </li>
--- a/index.bs
+++ b/index.bs
@@ -79,6 +79,9 @@ spec: BLUETOOTH-ASSIGNED
         text: Shortened Local Name; url: https://www.bluetooth.org/en-us/specification/assigned-numbers/generic-access-profile#
 
 spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
+    type: abstract-op
+        text: CanonicalNumericIndexString; url: sec-canonicalnumericindexstring
+        text: IsInteger; url: sec-isinteger
     type: dfn
         text: current realm; url: current-realm
         text: fulfilled; url: sec-promise-objects
@@ -86,6 +89,7 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
         text: realm; url: sec-code-realms
     type: method
         text: Array.prototype.map; url: sec-array.prototype.map
+        text: [[OwnPropertyKeys]]; for: Object; url: sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys
     type: interface
         text: Array; url: sec-array-objects
         text: ArrayBuffer; url: sec-arraybuffer-constructor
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -529,16 +533,22 @@ spec: html
   <h2 id="device-discovery" oldids="requestDevice-secure-context">Device Discovery</h2>
 
   <pre class="idl">
-    dictionary BluetoothRequestDeviceFilter {
+    dictionary BluetoothDataFilterInit {
+      BufferSource dataPrefix;
```
g-ortuno: This might be a silly question but what can you do with `dataPrefix` that you can't with just a `mask`? Is it that dataPrefix: [0x1] would match both service data: [0x1] and [0x3] but mask would only match the former?

index.bs
```diff
@@ -896,6 +964,32 @@ spec: html
         </td>
         <td>`namePrefix`, if present, must be non-empty to filter devices.</td>
       </tr>
+      <tr>
+        <td>
+          <pre highlight="js">
+            requestDevice({
+              filters:[{manufacturerData: {}}]
```
g-ortuno: Would it make sense to have this match devices that advertise any manufacturer data? Similarly for namePrefix: would it make sense for an empty string to match any device that advertises a name?

index.bs
jyasskin: If you want to match [0x1] and [0x3] but not [0x2], you'd use `{dataPrefix: [0x1], mask: [0xfd]}` to say that the lowest bit has to be 1, the second bit doesn't matter, and the rest have to be 0. There's no way to say that without using both `dataPrefix` and `mask`.

jyasskin: Should these take `BufferSource`s (ArrayBuffers and views) or sequences of bytes? Sequences of bytes will be more convenient to write out at the call site, but maybe harder to assemble programmatically if there are multi-byte fields in the data.

index.bs
```diff
@@ -804,28 +820,80 @@ spec: html
...
         </td>
         <td>D2</td>
       </tr>
       <tr>
         <td>
           <pre highlight="js">
-            [{manufacturerId: 17},
-             {serviceDataUUID: A}]
+            [{manufacturerData: {17: {}}},
+             {serviceData: {"A": {}}}]
```
jyasskin: Sure, done.

index.bs
```diff
@@ -529,16 +533,22 @@ spec: html
...
-      BluetoothServiceUUID serviceDataUUID;
+      // Maps unsigned shorts to BluetoothDataFilters.
+      object manufacturerData;
+      // Maps BluetoothServiceUUIDs to BluetoothDataFilters.
+      object serviceData;
     };
 
     dictionary RequestDeviceOptions {
-      sequence&lt;BluetoothRequestDeviceFilter> filters;
+      sequence&lt;BluetoothLEScanFilterInit> filters;
```
scheib: It's OK, but as a reader I didn't follow why these are FooTypeInit types instead of just FooType

index.bs
scheib: I bias slightly for acceptAllAdvertisements - not making it 'easy' to wildcard.

scanning.bs
```diff
@@ -131,7 +150,10 @@ spec:web-bluetooth-1; type:enum-value; for:PermissionName; text:"bluetooth"
         let myBeaconUuid = new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
 
         navigator.bluetooth.<a idl for="Bluetooth" lt="requestLEScan()">requestLEScan</a>({
-          filters: [{manufacturerId: 0x004C}],
+          filters: [{manufacturerData: {0x004C: {dataPrefix: new Uint8Array([
+            0x02, 0x15, // iBeacon identifier.
+            0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15  // My beacon UUID.
```
scheib: :+1:

scanning.bs
```diff
@@ -74,18 +74,37 @@ Markup Shorthands: css no, markdown yes
...
         text: Array.prototype.map; url: sec-array.prototype.map
+        text: [[OwnPropertyKeys]]; for: Object; url: sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys
     type: interface
         text: TypeError; url: sec-native-error-types-used-in-this-standard-typeerror
+
+spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/#
+    type: dfn
+        text: a copy of the bytes held; url: dfn-get-buffer-source-copy
+
+spec: web-bluetooth; urlPrefix: index.html#
```
jyasskin: These are in `anchors` here because we have to commit this PR for Shepherd to pick up the definitions in it.

index.bs
g-ortuno: Thanks for the explanation!

g-ortuno: I don't have a strong opinion either way but I think it might be a bit strange to use `BufferSource`s every where but in one place. Though I agree that it would make calls easier.

irgu:
At the beginning - thanks for decision that this feature will be included in specification. My only suggestion is that 'mask/prefix' is not so intuitive to use and something like 'pattern/ignore' would be little better and more flexible. I'm not javascript expert, so allowed myself to write small example in C to illustrate this:

```C
#include <stddef.h>
#include <stdbool.h>
#include <stdio.h>

char service_data_1[] = {0xAA, 0xB3, 0xCC}; //10101010 10110011 11001100
char service_data_2[] = {0xAE, 0x8B, 0xCD}; //10101110 10001011 11001101

char pattern[]        = {0xA0, 0x00, 0xCC}; //10100000 00000000 11001100
char ignore[]         = {0x0F, 0xFF, 0x00}; //00001111 11111111 00000000

bool service_data_check(char * p_service_data, size_t service_data_size, char * p_pattern, char * p_ignore)
{
    for(char i=0; i<service_data_size; i++)
    {
        if( (p_service_data[i] & (~ignore[i])) == (p_pattern[i] & (~ignore[i])) )
        {
            //byte matches - go ahead
        }
        else
        {
            //byte not matches - stop checking
            printf("Service data don't matches pattern!\n");
            return false;
        }
    }
    
    printf("Service data matches pattern!\n");
    return true;
}

int main()
{
    service_data_check(service_data_1, sizeof(service_data_1), pattern, ignore);
    service_data_check(service_data_2, sizeof(service_data_2), pattern, ignore);
}
```

jyasskin:
@irgu I think your proposal is to rename `dataPrefix` to `pattern` and rename `mask` to `~ignore`. Is that right? I'm definitely inclined to keep "prefix" in the name if the behavior is to assert a prefix. "Mask" vs "ignore" is a matter of taste; I've followed [Android's API](https://developer.android.com/reference/android/bluetooth/le/ScanFilter.html) here. Are there other Bluetooth APIs that use "ignore"?

irgu:
@jyasskin I looked at it once again and I totally agree with you. Current solution has the same flexibility and 'prefix' is reasonable name. It was just a small incomprehension of naming scheme when I saw this for the first time. Thanks for clearing :)

beaufortfrancois:
FYI I'm thinking of using `sequence` instead of `map` for those filters. See https://github.com/WebBluetoothCG/web-bluetooth/pull/545

PR #301, opened 2016-09-28
Record the Chrome implementation status of the whole GATT API.

jyasskin:
I didn't break the Scanning API down into components yet.

@g-ortuno @beaufortfrancois

### DIFF
--- a/implementation-status.md
+++ b/implementation-status.md
@@ -11,29 +11,36 @@ Work is in progress:
 * The `chrome://flags/#enable-web-bluetooth` flag must be enabled, or the web page must have an [origin trial meta tag or header](https://github.com/jpchase/OriginTrials/blob/gh-pages/developer-guide.md) with a token requested from [http://bit.ly/WebBluetoothOriginTrial](http://bit.ly/WebBluetoothOriginTrial).
 * Root [Issue 419413: Web Bluetooth](https://code.google.com/p/chromium/issues/detail?id=419413) and blocking issues are most authorative on status.
 
+## [GATT Communication API](https://webbluetoothcg.github.io/web-bluetooth/)
+
 Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows | iOS
 ------------------------- | :-------: | :-----: | :-: | :---: | :-----: | :-:
 getAvailability()         |           |         |     |       |         |
-Discovery                 | ✓         | ✓       | ✓   | ✓     | ✓       | See notes
+Referring Device (Physical Web) |     |         |     |       |         |
+Discovery                 | ✓         | ✓       | ✓   | ✓     | ✓       |
 └ Name or prefix          | ✓         | ✓       | ✓   | ✓     | ✓       |
 └ Manufacturer/Service data |         |         |     |       |         |
 └ acceptAllDevices        |           |         |     |       |         |
-Referring Device (Physical Web) |     |         |     |       |         |
 Chooser UI                | ✓         | ✓       | ✓   | ✓     | ✓       |
+Permissions API Integration |         |         |     |       |         |
 GATT Server Connect       | ✓         | ✓       | ✓   | ✓     |         |
-Read Characteristic       | ✓         | ✓       | ✓   | ✓     |         |
-Write Characteristic      | ✓         | ✓       | ✓   | ✓     |         |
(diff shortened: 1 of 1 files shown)


implementation-status.md
```diff
@@ -11,29 +11,48 @@ Work is in progress:
...
-Referring Device (Physical Web) |     |         |     |       |         |
 Chooser UI                | ✓         | ✓       | ✓   | ✓     | ✓       |
+permissions.request()     |           |         |     |       |         |
+permissions.query()       |           |         |     |       |         |
+permissions.revoke()      |           |         |     |       |         |
+watchAdvertisements()     |           |         |     |       |         |
+Persistent Device IDs     |           |         |     |       |         |
 GATT Server Connect       | ✓         | ✓       | ✓   | ✓     |         |
+GATT Server Disconnect    | ✓         | ✓       | ✓   | ✓     |         |
+getPrimaryService*()      | ✓         | ✓      | ✓   | ✓     |         |
```
scheib: There's a space/tab mismatch here.

implementation-status.md
```diff
@@ -11,29 +11,48 @@ Work is in progress:
 * The `chrome://flags/#enable-web-bluetooth` flag must be enabled, or the web page must have an [origin trial meta tag or header](https://github.com/jpchase/OriginTrials/blob/gh-pages/developer-guide.md) with a token requested from [http://bit.ly/WebBluetoothOriginTrial](http://bit.ly/WebBluetoothOriginTrial).
 * Root [Issue 419413: Web Bluetooth](https://code.google.com/p/chromium/issues/detail?id=419413) and blocking issues are most authorative on status.
 
+## [GATT Communication API](https://webbluetoothcg.github.io/web-bluetooth/)
+
 Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows | iOS
 ------------------------- | :-------: | :-----: | :-: | :---: | :-----: | :-:
 getAvailability()         |           |         |     |       |         |
+Referring Device (Physical Web) |     |         |     |       |         |
 Discovery                 | ✓         | ✓       | ✓   | ✓     | ✓       | See notes
```
beaufortfrancois: Shall we move "See notes" somewhere else then?

implementation-status.md
```diff
@@ -11,29 +11,48 @@ Work is in progress:
...
 Write Characteristic      | ✓         | ✓       | ✓   | ✓     |         |
-Characteristic Properties | ✓         | ✓       | ✓   | ✓     |         |
 Start GATT Notifications  | ✓         | ✓       | ✓   | ✓     |         |
 Stop GATT Notifications   | ✓         | 54      |     | ✓     |         |
-GATT Server Disconnect    | ✓         | ✓       | ✓   | ✓     |         |
-Get Characteristics List  | ✓         | ✓       | ✓   | ✓     |         |
-Device Disconnected Event | ✓         | ✓       | ✓   | ✓     |         |
-Get Primary Services List | ✓         | ✓      | ✓   | ✓     |         |
 {start,stop}Notifications returns `this` | 54 | 54 | 54 | 54  |         |
+Descriptors               |           |         |     |       |         |
```
beaufortfrancois: Should we split this in "Read Descriptor" and "Write Descriptor"?

implementation-status.md
jyasskin: Maybe. Do you have an idea for the best place to put it? On getAvailability() because that's the first line of the table, or outside of the table because it applies to the whole thing?

implementation-status.md
jyasskin: I was thinking we'd only need to split this if descriptors land in multiple chunks. We could merge some of the characteristic features too, to shorten the table.

implementation-status.md
jyasskin: Fixed.

implementation-status.md
beaufortfrancois: I would simply remove it from this table.

implementation-status.md
beaufortfrancois: Makes sense.

I think we will need to wait for `stopNotifications` to work on macOS to be able to merge all in one simple row "Characteristics". WDYT?

jyasskin: SG.

implementation-status.md
jyasskin: Done.

PR #302, opened 2016-09-29
Add Android Webview column

beaufortfrancois:
R: @jyasskin  @scheib


implementation-status.md
```diff
@@ -11,8 +11,8 @@ Work is in progress:
 * The `chrome://flags/#enable-web-bluetooth` flag must be enabled, or the web page must have an [origin trial meta tag or header](https://github.com/jpchase/OriginTrials/blob/gh-pages/developer-guide.md) with a token requested from [http://bit.ly/WebBluetoothOriginTrial](http://bit.ly/WebBluetoothOriginTrial).
 * Root [Issue 419413: Web Bluetooth](https://code.google.com/p/chromium/issues/detail?id=419413) and blocking issues are most authorative on status.
 
-Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows | iOS
-------------------------- | :-------: | :-----: | :-: | :---: | :-----: | :-:
+Feature/Platform          | Chrome OS | Android | Mac | Linux | Windows | iOS | Android WebView
```
jyasskin: This is fine, but I wonder if it would be clearer to put iOS and Android WebView into a different table where we say explicitly that they're not supported at all yet. With them in this table, folks have to search their column for a ✓ to figure out if we support them at all.

beaufortfrancois: You're right. I've updated this CL.
Let me know what you think.

PR #308, opened 2016-10-07
Mark sections that are more likely to change.

jyasskin:
These are all unimplemented, and haven't been thought through as well as
the rest of the spec.

Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/at-risk/index.bs#permission-api-integration.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2378,7 +2378,7 @@ spec: html
       and should call {{unwatchAdvertisements()}} to stop using power as soon as possible.
     </p>
 
-    <div algorithm>
+    <div algorithm class="unstable">
       <p>
         The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code> method,
         when invoked,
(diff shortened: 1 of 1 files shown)


g-ortuno:
What about advertisements?

jyasskin:
I think the `advertisementreceived` event is pretty stable even though we haven't implemented it; do you think it's likely to change?

scheib:
I'd err on the conservative side and mark it not stable if no implementation has started on it.

jyasskin:
Ok, done. Do you think getIncludedService(), descriptors, referringDevice, acceptAllDevices, and Manufacturer/Service data are stable enough even though we haven't started implementing them?

jyasskin:
Ping. I also marked `referringDevice` as unstable, given our difficulties in getting the MAC address from Android.

scheib:
Thanks, I'd use the rule of "if no one has implemented it - then it is unstable". It is the most conservative thing to do, but also sends a clear message about what is easily malleable in the spec.

jyasskin:
Done.

PR #310, opened 2016-10-11
Add "invalidate all objects when a device disconnects" row

beaufortfrancois:
Background at  https://bugs.chromium.org/p/chromium/issues/detail?id=654687


implementation-status.md
```diff
@@ -45,6 +45,7 @@ Device Disconnected Event | ✓         | ✓       | ✓   | ✓     |
 Service Changed Event     |           |         |     |       |         |
 BluetoothUUID             | ✓         | ✓       | ✓   | ✓     | ✓       |
 TypeError for bad UUIDs   | 55        | 55      | 55  | 55    | 55      |
+Invalidate objects upon disconnect |  |         |     |       |         |
```
jyasskin: Hm, from https://crbug.com/654687, it looks like this is working on MacOS?

beaufortfrancois:
Shall we merge as it and only add its official support when https://bugs.chromium.org/p/chromium/issues/detail?id=654950 is fixed?

jyasskin:
Sure.

Issue #188, opened 2015-11-24
GATT operation in progress - how to handle it?


perja12:
I have a web page that writes to a characteristic, but if I do this too fast, I will get the error "GATT operation in progress". Is this something the page author should handle or should the implementation take care of queuing the requests?

If it is the page author, it will be good to include some recommendations for how to handle this. AFAIK there is no way to query the state of the implementation other than responding to the errors.

beaufortfrancois:
Are you testing this on Chrome OS or Android?

perja12:
Testing this on Android (Chrome Dev  build) and soon in Opera internal build.

beaufortfrancois:
I believe this issue has been fixed recently (https://codereview.chromium.org/1465863003/).
You may want to latest Chromium build at https://download-chromium.appspot.com/?platform=Android and let me know if you're still experiencing this.

perja12:
Tested the latest Chromium and the crash when "GATT operation in progress" occurs has been fixed - so that is good. 

But my main question is still open: is it the responsibility of the page author to handle "GATT operation in progress" or should this be handled by the implementation? There seems to be pros and cons with both approaches.

beaufortfrancois:
Glad to hear that.

Hopefully @g-ortuno @scheib and @jyasskin will be able to fully answer your question here.

jyasskin:
The implementation's supposed to handle it, but we don't yet (https://crbug.com/560862). @scheib has argued that we should make the web page handle it instead, in which case the rule would be that you wait for the `writeValue()` Promise to settle before calling `writeValue()` again. If there's a good reason you can't do that, it would probably settle the argument.

scheib:
We should keep the implementation flexible, and not force queuing if the application doesn't desire it.

12:35 AM <perja> Hi, short question about web-bluetooth: I have a page that writes to a characteristic whenever a button is pressed (or actually a touch event), but if I click to fast I get the  "GATT operation in progress"-error. Is it so that the page author is supposed to handle this with a queue or similar, or will the implementation do the queing?
1:12 AM <scheib> perja: Currently unspecified. We should pick and make it clear. Some devices may be fairly slow to accept a write. What should happen if the web page can generate write requests at 2 times the speed a device can process them? If the browser queues, should it pick any limit for how large the queue should get? An application can know if a queue of unlimited values is needed or if only 1 pending value needs to be queued and replaced if a newer value arrives.

pzboyz:
[...]

PR #316, opened 2016-10-20
Allow UAs to reject concurrent use of 6 functions.

jyasskin:
Preview at https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/web-bluetooth-1/in-progress-error/index.bs#dom-bluetoothremotegattserver-connect.

* GattServer.connect()
* Characteristic.readValue()
* Characteristic.writeValue()
* Characteristic.startNotifications()
* Descriptor.readValue()
* Descriptor.writeValue()

This is uncomfortable because while serializing within a single realm
will reduce the frequency of these NetworkErrors, separate realms and
possibly other applications entirely can also cause these errors, and
the only way to recover is to retry and hope it doesn't happen again.

We're discussing the constraints imposed by the OS and the long-term
plan in #188. Any change should reduce the number of exceptions and so
be backward-compatible.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3231,6 +3231,16 @@ spec: html
           and abort these steps.
         </li>
         <li>
+          If the UA is currently using the Bluetooth system,
+          it MAY <a>queue a task</a> to <a>reject</a> |promise| with a {{NetworkError}},
+          return |promise|,
+          and abort these steps.
+
+          Issue(188): Implementations may be able to avoid this {{NetworkError}},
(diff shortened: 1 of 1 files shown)


jyasskin:
I'm going to merge this based on https://github.com/WebBluetoothCG/web-bluetooth/issues/188#issuecomment-255319483, but feel free to refine it as y'all discover more concurrency that works everywhere.

PR #319, opened 2016-10-26
Create use-cases.html

anirudhcmohan:
Fix use-cases.html to also be on gh-pages branch.


g-ortuno:
fwiw: The proper way to fix this would be to turn use-cases into a .bs file. Otherwise they will get out of sync very easily.

PR #323, opened 2016-10-27
Fix bikeshed #851 issue

beaufortfrancois:
It fixes https://github.com/tabatkins/bikeshed/issues/851

See https://api.csswg.org/bikeshed/?url=https://github.com/WebBluetoothCG/web-bluetooth/raw/fixBSDfn/index.bs

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1084,7 +1084,7 @@ spec: html
   <div algorithm="initializing referringDevice">
     <p>
       Some UAs may allow the user
-      to cause a <a>browsing context</a> to <a lt="navigated">navigate</a>
+      to cause a <a for="/">browsing context</a> to <a lt="navigated">navigate</a>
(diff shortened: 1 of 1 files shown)


scheib:
It seems https://github.com/tabatkins/bikeshed/issues/851 fixes this in bikeshed. Shouldn't the fix in our spec simply be to re-process using the updated bikeshed?

g-ortuno:
Rather than adding this to each link we should add `text: browsing context; for: /` to line [118](https://github.com/WebBluetoothCG/web-bluetooth/blob/master/index.bs#L118)

@scheib I think tabatkins/bikeshed#851 fixes the error message (it was missing the `for: /` part).

g-ortuno:
Closing now that the fix landed in #328 @beaufortfrancois: Thanks for finding the issue!

PR #328, opened 2016-11-01
Disambiguate links for browsing context

g-ortuno:
Preview at: https://api.csswg.org/bikeshed/?url=https://github.com/g-ortuno/web-bluetooth/raw/add-browsing-context-default/index.bs


g-ortuno:
@scheib PTAL. @beaufortfrancois is OOO and we need this to be able to land the use-cases change.

PR #332, opened 2016-11-03
Add use-cases to the list of specs.

g-ortuno:
That way deploy.sh generates use-cases.html


g-ortuno:
@scheib PTAL :sweat_smile:

PR #338, opened 2016-12-06
Fix typo in GetPrimaryServices

dati91:
I think there is a typo in `GetPrimaryServices`, because `this{.device}` is used everywhere else but here.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/dati91/web-bluetooth/raw/gattchildren_typo/index.bs


g-ortuno:
Thanks!

Issue #326, opened 2016-11-01
Should derive from EventTarget, not implement it


domenic:
Implements should be used for mixins, but EventTarget is not a mixin.

PR #344, opened 2016-12-26
Let Bluetooth inherit from EventTarget

foolip:
This is the same problem as in https://github.com/WICG/webusb/pull/67

Here, too, sprinkle some [SameObject] around.


zcorpan:
This fixes https://github.com/WebBluetoothCG/web-bluetooth/issues/326

PR #349, opened 2017-01-19
Fixes for the Permission API integration part.

### LINKED ISSUE #347: Use the canonicalizing algorithm when querying the bluetooth permission
In the [query the "bluetooth" permission algorithm](https://webbluetoothcg.github.io/web-bluetooth/#query-the-bluetooth-permission) before `Step 6.2` I think there should be a check with calling the [canonicalization](https://webbluetoothcg.github.io/web-bluetooth/#bluetoothlescanfilterinit-canonicalizing) on each filter in `desc.filters`, to prevent using not valid filters or passing blocklisted services. If any of these calls return with an error, propagate that error and abort the steps. Otherwise use the results of the canonicalization to match a filter in `Step 6.2`.

zakorgy:
This PR fixes #347, also removes the numbering from the Note element in the query algorithm.

Preview at https://api.csswg.org/bikeshed/?url=https://github.com/zakorgy/web-bluetooth/raw/permission-request/index.bs#query-the-bluetooth-permission.

---
This change is [Reviewable](https://reviewable.io/reviews/webbluetoothcg/web-bluetooth/349)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1947,24 +1947,13 @@ spec: html
                 If <code><var>desc</var>.filters</code> is set, do the following sub-steps:
                 <ol>
                   <li>
-                    Let <var>uuidFilters</var> be a new {{Array}},
-                  </li>
-                  <li>
-                    For each |filter| in <code><var>desc</var>.filters</code>, do the following sub-steps:
-                    <ol>
-                      <li>
-                        Let |canonicalFilter| be
-                        the result of <a for="BluetoothLEScanFilterInit">canonicalizing</a> |filter|.
-                        If this step returns an error, return that error and abort these steps.
-                      </li>
-                      <li>
-                        Append |canonicalFilter| to |uuidFilters|.
-                      </li>
-                    </ol>
+                    Replace each filter in <code>|desc|.filters</code>
+                    with the result of <a for="BluetoothLEScanFilterInit">canonicalizing</a> it.
+                    If any of these canonicalizations throws an error, return that error and abort these steps.
                   </li>
                   <li>
                     If <code><var>allowedDevice</var>.{{[[device]]}}</code> does not
-                    <a>match a filter</a> in <var>uuidFilters</var>,
+                    <a>match a filter</a> in <code>|desc|.filters</code>,
                     continue to the next <var>allowedDevice</var>.
                   </li>
                 </ol>
--- a/index.bs
+++ b/index.bs
@@ -1944,14 +1944,33 @@ spec: html
                 continue to the next <var>allowedDevice</var>.
               </li>
               <li>
-                If <code><var>desc</var>.filters</code> is set
-                and <code><var>allowedDevice</var>.{{[[device]]}}</code> does not
-                <a>match a filter</a> in <code><var>desc</var>.filters</code>,
-                continue to the next <var>allowedDevice</var>.
+                If <code><var>desc</var>.filters</code> is set, do the following sub-steps:
+                <ol>
+                  <li>
+                    Let <var>uuidFilters</var> be a new {{Array}},
+                  </li>
+                  <li>
+                    For each |filter| in <code><var>desc</var>.filters</code>, do the following sub-steps:
+                    <ol>
+                      <li>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1944,14 +1944,33 @@ spec: html
...
+                  <li>
+                    Let <var>uuidFilters</var> be a new {{Array}},
+                  </li>
+                  <li>
+                    For each |filter| in <code><var>desc</var>.filters</code>, do the following sub-steps:
+                    <ol>
+                      <li>
+                        Let |canonicalFilter| be
+                        the result of <a for="BluetoothLEScanFilterInit">canonicalizing</a> |filter|.
+                        If this step returns an error, return that error and abort these steps.
```
jyasskin: Interestingly, the [permission query algorithm](https://w3c.github.io/permissions/#permission-query-algorithm) can't return an error. That's a mistake, which I've filed as w3c/permissions#135.

index.bs
```diff
@@ -1875,7 +1875,8 @@ spec: html
       </dt>
       <dd algorithm="permission request">
         <p>
-          Given a {{BluetoothPermissionDescriptor}} <var>options</var>
+          To <dfn>request the "bluetooth" permission</dfn> with
```
jyasskin: Could you send mutually-independent changes as separate PRs?

index.bs
```diff
@@ -1944,14 +1945,33 @@ spec: html
...
-                and <code><var>allowedDevice</var>.{{[[device]]}}</code> does not
-                <a>match a filter</a> in <code><var>desc</var>.filters</code>,
-                continue to the next <var>allowedDevice</var>.
+                If <code><var>desc</var>.filters</code> is set, do the following sub-steps:
+                <ol>
+                  <li>
+                    Let <var>uuidFilters</var> be a new {{Array}},
+                  </li>
+                  <li>
+                    For each |filter| in <code><var>desc</var>.filters</code>, do the following sub-steps:
```
jyasskin: I think we can safely write this as `Replace each filter in <code>|desc|.filters</code> with the result of <a for="BluetoothLEScanFilterInit">canonicalizing</a> it. If any of these canonicalizations throws an error ...` since *desc* is a dictionary instead of a reference to an object argument.

index.bs
zakorgy: I created a different PR for this https://github.com/WebBluetoothCG/web-bluetooth/pull/353

index.bs
```diff
@@ -1944,14 +1944,22 @@ spec: html
...
+                  </li>
+                  <li>
+                    If <code><var>allowedDevice</var>.{{[[device]]}}</code> does not
+                    <a>match a filter</a> in <code>|desc|.filters</code>,
+                    continue to the next <var>allowedDevice</var>.
+                  </li>
+                </ol>
               </li>
-              <li class="note">
+
```
jyasskin: This results in the Note living inside the previous `<li>`, which is about `desc.filters` rather than `desc.optionalServices`. Could you leave the `<li>` around it?

zakorgy: Putting back the `<li>` even with `class="note"`adds the numbering for the Note part, but this is not a step, so there should be no number before it. Other solution I think is to move the Note part after the `</ol>` in [line 1969](https://github.com/WebBluetoothCG/web-bluetooth/pull/349/files/745d2518f7a267f18de1abd9ad71919519170f23#diff-ec9cfa5f3f35ec1f84feb2e59686c34dR1969).

zakorgy:
@jyasskin  I came up with a different idea, which I mentioned [here](https://github.com/WebBluetoothCG/web-bluetooth/pull/349#discussion_r99774324). If you doesn't like it, I will replace the `</li>`.

index.bs
jyasskin: That's fine. The purpose of a specification is to help multiple organizations build implementations that behave the same way, and secondarily to help developers use those implementations. The presence or absence of a number for a note doesn't affect that, so it's generally not worth spending much time on.

jyasskin:
Also, thanks for the fix!

PR #368, opened 2017-06-30
WEB Bluetooth

ivanaxel:
luetooth is a standard for short-range wireless communication between devices. This group is developing a specification for Bluetooth APIs to allow websites to communicate with devices in a secure and privacy-preserving way. In particular the web Bluetooth API focuses on minimizing the device attack surface exposed to malicious websites, possibly by removing access to some existing Bluetooth features that are hard to implement securely. Further, the API takes the approach of a user interface to select and approve access to devices as opposed to using certification and installation.

Most of our activity happens in our GitHub repository, with supporting code in adjacent repositories in the WebBluetoothCG GitHub organization.

Note: Community Groups are proposed and run by the community. Although W3C hosts these conversations, the groups do not necessarily represent the views of the W3C Membership or staff.


beaufortfrancois:
Closing this

PR #389, opened 2018-04-02
Use `interface mixins` instead of `[NoInterfaceObject]`

romandev:
WebIDL recently introduced dedicated syntax for mixins[1]. So, we can
replace `[NoInterfaceObject]` and `implements` with `interface mixin` and
`includes`.

[1] https://github.com/heycam/webidl/commit/45e8173d40ddff8dcf81697326e094bcf8b92920


romandev:
Hmm.. I missed that `interface mixin` can't have writable attributes.
I'll check it first and then re-upload later.

PR #391, opened 2018-04-12
Handle notifications for characteristics missing a CCCD

baileyforrest:
A common violation of the specification is characteristics with
notification/indication capabilities not having an associated
Client Characteristic Configuration descriptor (CCCD).

In order to allow compatability with these types of devices,
implementaions should still attempt to register notifications on
characteristics without a CCD.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4044,7 +4044,8 @@ spec: html
           and/or give the user a way to retry failed operations.
         </li>
         <li>
-          Use any of the <a>Characteristic Descriptors</a> procedures
+          If the characteristic has a <a>Client Characteristic Configuration</a>
+          descriptor, use any of the <a>Characteristic Descriptors</a> procedures
           to ensure that one of the <code>Notification</code> or <code>Indication</code> bits in
           <var>characteristic</var>'s <a>Client Characteristic Configuration</a> descriptor
           is set, matching the constraints
@@ -4053,6 +4054,13 @@ spec: html
           and MUST deduplicate <a href="#notification-events">value-change events</a>
           if both bits are set.
           Handle errors as described in <a href="#error-handling"></a>.
+
+          Note: Some devices have characteristics whose properties include the
+          Notify or Indicate bit but that don't have a <a>Client Characteristic
+          Configuration</a> descriptor. These non-standard-compliant characteristics
+          tend to send notifications or indications unconditionally, so this
+          specification allows applications to simply subscribe to their
+          messages.
         </li>
         <li>
           If the previous step returned an error,
@@ -4099,7 +4107,8 @@ spec: html
           remove it.
         </li>
         <li>
-          If <var>characteristic</var>'s <a>active notification context set</a> became empty,
+          If <var>characteristic</var>'s <a>active notification context set</a> became empty
+          and the characteristic has a <a>Client Characteristic Configuration</a> descriptor,
           the UA SHOULD use any of the <a>Characteristic Descriptors</a> procedures
           to clear the <code>Notification</code> and <code>Indication</code> bits in
           <var>characteristic</var>'s <a>Client Characteristic Configuration</a> descriptor.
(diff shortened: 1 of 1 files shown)


baileyforrest:
Preview URL:
https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/baileyforrest/web-bluetooth/master/index.bs

beaufortfrancois:
@jyasskin What do you think of this proposal?

baileyforrest:
There's some more context here:
https://chromium-review.googlesource.com/c/chromium/src/+/1008984

jyasskin:
@beaufortfrancois, have you seen other reports of devices that send notifications and indications but don't expose a CCCD to let their clients configure that? I'm reluctant to diverge from the specification, but if this mistake is common, we might have to.

If we do this, the specification should be phrased like

> If the characteristic has a Client Characteristic Configuration descriptor, use any of the Characteristic Descriptors procedures to ensure that one of the Notification or Indication bits in characteristic’s Client Characteristic Configuration descriptor is set, matching the constraints in characteristic’s properties. The UA SHOULD avoid setting both bits, and MUST deduplicate value-change events if both bits are set. Handle errors as described in §5.7 Error handling.
>
> Note: Some devices have characteristics whose properties include the Notify or Indicate bit but that don't have a Client Characteristic Configuration descriptor. These non-standard-compliant characteristics tend to send notifications or indications unconditionally, so this specification allows applications to simply subscribe to their messages.

That is, the behavior for non-compliant devices should be included in the main flow of the algorithm instead of [COMEFROM](https://en.wikipedia.org/wiki/COMEFROM)ed later.

baileyforrest:
I'm not sure exactly how common this error is, but I see that it's very easy to make this mistake on many BLE GATT server APIs including Android's API.

My worry is if the standard doesn't support this case, then Web BT won't work with many types of devices.

Emill:
I wouldn't believe the mistake is that common. I mean for example all Android client tutorials online that deal with notifications show how to write the descriptor, so if people implement their peripheral wrong, they will notice that the descriptor is missing anyway if they connect to it from Android.

And peripherals set up using Android's GATT server api can easily be fixed by releasing app updates.

If this workaround is added, I suspect it rather can increase the amount of buggy devices, since they won't notice that they have done wrong. It even encourages this breaking of the rules since it would save one roundtrip since you don't need to write the descriptor then...

reillyeon:
@baileyforrest Given your understanding of how this mistake is made does @jyasskin's proposed language produce the correct behavior?

reillyeon:
@Emill, to mitigate this perhaps implementations could emit a console warning when the CCCD is not present.

jyasskin:
It sounds like Android and iOS can both receive notifications from broken devices like the one described here. Like @Emill I wouldn't want Web Bluetooth to be the only one that allows them, but since we're so much smaller than the existing ecosystem, we 1) probably don't have a choice, and 2) probably won't make things any worse than they already are.

baileyforrest:
I updated the pull request based on @jyasskin's suggestion. PTAL

index.bs
```diff
@@ -4099,10 +4107,18 @@ spec: html
...
         </li>
         <li>
-          If <var>characteristic</var>'s <a>active notification context set</a> became empty,
+          If <var>characteristic</var>'s <a>active notification context set</a> became empty
+          and the characteristic has a <a>Client Characteristic Configuration</a> descriptor,
           the UA SHOULD use any of the <a>Characteristic Descriptors</a> procedures
           to clear the <code>Notification</code> and <code>Indication</code> bits in
           <var>characteristic</var>'s <a>Client Characteristic Configuration</a> descriptor.
+
+          Note: Some devices have characteristics whose properties include the
```
jyasskin: I think we don't need the Note here: we allow folks to `stopNotifications()` even when no notifications are enabled, and even when the Notify or Indicate bit isn't set, and a missing CCCD is just a corner case we need to handle regardless of the policy in `startNotifications()`.

baileyforrest: Done.

index.bs
```diff
@@ -4099,7 +4107,8 @@ spec: html
           remove it.
         </li>
         <li>
-          If <var>characteristic</var>'s <a>active notification context set</a> became empty,
+          If <var>characteristic</var>'s <a>active notification context set</a> became empty
+          and the characteristic has a <a>Client Characteristic Configuration</a> descriptor,
```
g-ortuno: We should add this to the (Clean up disconnected devices algorithm)[https://webbluetoothcg.github.io/web-bluetooth/#clean-up-the-disconnected-device].

baileyforrest: Done

PR #396, opened 2018-05-09
Replace "non-`null`" with "not `null`"

reillyeon:
Bikeshed compilation fails when encountering the former with,

"Unknown inline tag: non-`null`"

Replacing this with "not `null`" does not change the meaning.

### DIFF
--- a/scanning.bs
+++ b/scanning.bs
@@ -759,19 +759,19 @@ spec:web-bluetooth
       </p>
       <ul>
         <li>
-          If <code>|filter|.name</code> is non-`null`,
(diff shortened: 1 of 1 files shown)


reillyeon:
Tab acknowledges that this was caused by a change to Bikeshed. They suggested I replace the text with non-<code>null</code> instead and updated this pull request to do so.

tabatkins:
It's caused by a new (currently undocumented while it's experimental) feature that makes Bikeshed more extensible by outside scripts, by making ``foo`bar` `` (tagged code spans) a special thing.

PR #401, opened 2018-07-03
Add [SecureContext] to navigator.bluetooth

beaufortfrancois:
It is tricky sometimes to know why the Web Bluetooth API is not "working" as `navigator.bluetooth` is available in HTTP, but `navigator.bluetooth.requestDevice` only in HTTPS. I propose adding `[SecureContext]` to `navigator.bluetooth` as well.

WDYT @jyasskin 

CC @g-ortuno

***
<a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/pull/401.html" title="Last updated on Jul 6, 2018, 4:11 PM GMT (e7a3afb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/401/68c5a71...e7a3afb.html" title="Last updated on Jul 6, 2018, 4:11 PM GMT (e7a3afb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5355,8 +5355,9 @@ spec: promises-guide-1
   <h2 id="navigator-extensions">Extensions to the Navigator Interface</h2>
 
   <pre class="idl">
+    [SecureContext]
     partial interface Navigator {
-      [SameObject, SecureContext]
+      [SameObject]
       readonly attribute Bluetooth bluetooth;
     };
   </pre>
--- a/index.bs
+++ b/index.bs
@@ -5359,7 +5359,7 @@ spec: promises-guide-1
 
   <pre class="idl">
     partial interface Navigator {
-      [SameObject]
+      [SameObject, SecureContext]
       readonly attribute Bluetooth bluetooth;
     };
   </pre>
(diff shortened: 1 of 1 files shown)


jyasskin:
@reillyeon, any concerns about merging this into Chromium?

reillyeon:
I don't. I think that when this was first specced out [SecureContext] was new and there was some concern that it would be confusing to have the attribute simply missing in some contexts and we wanted to put up a console error instead. It appears that this code has since been removed so rather than delay until requestDevice() is called we should just hide everything behind [SecureContext].

index.bs
```diff
@@ -5359,7 +5359,7 @@ spec: promises-guide-1
 
   <pre class="idl">
     partial interface Navigator {
-      [SameObject]
```
reillyeon: Move the [SecureContext] annotation to apply to the whole partial Navigator interface and remove it from the attributes on the Bluetooth interface.

beaufortfrancois: Done.

reillyeon:
@jyasskin do you think we need to send an I2I+S to land this in Blink?

jyasskin:
@reillyeon Ask a Blink API owner. My guess would be no.

PR #416, opened 2018-11-19
Remove unused references from scanning.bs

reillyeon:


reillyeon:
FYI @tabatkins I'm not sure that the tests.bs change here is correct as it means that [[BLUETOOTH-ASSIGNED]] is not defined but is used in the anchors section.

tabatkins:
Right, given that your anchors block is asserting the existing of a BLUETOOTH-ASSIGNED spec (and you do actually use those anchors in the spec), you should have a biblio entry for it so it can show up in the bibliography properly.

The only failure right now is the lack of an entry in your biblio index (all your links work fine otherwise), but still.

reillyeon:
The problem is that bikeshed complains about the unused biblio entry when the only reference is in the anchors. Is this a bug or do we need to do something additional to get the biblio entry to appear in the spec other than use anchors that reference it?

tabatkins:
Ah, that's a bug then. ^_^

PR #419, opened 2018-12-10
Drop |name| attribute since it is accessible from the |device| attribute

dougt:
***
<a href="https://pr-preview.s3.amazonaws.com/dougt/web-bluetooth/pull/419.html" title="Last updated on Dec 10, 2018, 6:59 PM GMT (957ff9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/419/742618d...dougt:957ff9a.html" title="Last updated on Dec 10, 2018, 6:59 PM GMT (957ff9a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2496,7 +2496,6 @@ spec: promises-guide-1
           [SameObject]
           readonly attribute BluetoothDevice device;
           readonly attribute FrozenArray&lt;UUID> uuids;
-          readonly attribute DOMString? name;
           readonly attribute unsigned short? appearance;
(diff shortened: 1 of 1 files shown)


reillyeon:
I think there's a decision to be made here about whether or not attributes like `name` in the advertising event itself should reflect whether or not the attribute was in that particular advertising packet.

Emill:
The advertised name can be different than the name stored in the GATT db. The advertised name could also change for every advertisement.

PR #421, opened 2018-12-13
This makes all but the device and uuids attributes optional

dougt:
In both BluetoothAdvertisingEvent and BluetoothAdvertisingEventInit, there are a number of attributes that should be empty since underlying platform advertising events might not contain those fields.

I am not sure if uuids sequence should be optional.

***
<a href="https://pr-preview.s3.amazonaws.com/dougt/web-bluetooth/pull/421.html" title="Last updated on Dec 13, 2018, 10:09 PM GMT (a3ed290)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/421/aaf7b68...dougt:a3ed290.html" title="Last updated on Dec 13, 2018, 10:09 PM GMT (a3ed290)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2501,19 +2501,19 @@ spec: promises-guide-1
           readonly attribute byte? txPower;
           readonly attribute byte? rssi;
           [SameObject]
-          readonly attribute BluetoothManufacturerDataMap manufacturerData;
+          readonly attribute BluetoothManufacturerDataMap? manufacturerData;
           [SameObject]
-          readonly attribute BluetoothServiceDataMap serviceData;
+          readonly attribute BluetoothServiceDataMap? serviceData;
         };
         dictionary BluetoothAdvertisingEventInit : EventInit {
           required BluetoothDevice device;
           sequence&lt;(DOMString or unsigned long)> uuids;
-          DOMString name;
-          unsigned short appearance;
-          byte txPower;
-          byte rssi;
-          BluetoothManufacturerDataMap manufacturerData;
-          BluetoothServiceDataMap serviceData;
+          DOMString? name;
+          unsigned short? appearance;
+          byte? txPower;
+          byte? rssi;
+          BluetoothManufacturerDataMap? manufacturerData;
+          BluetoothServiceDataMap? serviceData;
         };
       </pre>
 


g-ortuno:
According to Core Specification Supplement V7 Part A 1 DATA TYPES DEFINITIONS AND FORMATS, all fields are optional so I think it would make sense for `uuids` to also be optional. Note that RSSI is not part of the advertisement packet itself, and should not be optional.

g-ortuno:
Oh interesting. I was assuming they were optional because they are optional in the actual advertisement packet, not because some platforms might not expose them.

index.bs
```diff
@@ -2501,19 +2501,19 @@ spec: promises-guide-1
           readonly attribute byte? txPower;
           readonly attribute byte? rssi;
           [SameObject]
-          readonly attribute BluetoothManufacturerDataMap manufacturerData;
+          readonly attribute BluetoothManufacturerDataMap? manufacturerData;
```
jyasskin: The maps weren't optional because they'll just be empty if they're not sent. I haven't double-checked today if the advertising packet has distinct not-present and empty states? If not, I don't want the returned object to provide distinct states for those either.

index.bs
```diff
@@ -2501,19 +2501,19 @@ spec: promises-guide-1
...
           [SameObject]
-          readonly attribute BluetoothServiceDataMap serviceData;
+          readonly attribute BluetoothServiceDataMap? serviceData;
         };
         dictionary BluetoothAdvertisingEventInit : EventInit {
           required BluetoothDevice device;
           sequence&lt;(DOMString or unsigned long)> uuids;
```
jyasskin: Similarly with uuids, if none are included in the ad packet, this list should just be empty, rather than null.

index.bs
```diff
@@ -2501,19 +2501,19 @@ spec: promises-guide-1
...
-          DOMString name;
-          unsigned short appearance;
-          byte txPower;
-          byte rssi;
-          BluetoothManufacturerDataMap manufacturerData;
-          BluetoothServiceDataMap serviceData;
+          DOMString? name;
```
jyasskin: Fields in a dictionary are [optional by default](https://heycam.github.io/webidl/#dfn-present). They only become required if the [`required` keyword](https://heycam.github.io/webidl/#required-dictionary-member) is applied.

dougt:
Thank you for that explanation!

g-ortuno:
All the fields in the advertisement packet are optional; why can't the ones in the API be optional as well?

jyasskin:
@g-ortuno I think they all are optional today. If there isn't any manufacturer data, the `manufacturerData` map has no elements. If there are no UUIDs, the `uuids` array is empty. If there's no name, the `name` is null.

index.bs
reillyeon: To extend that: An optional field can take the value `undefined`. Adding the "?" to the type means that it can also take the value `null` which is unnecessary in this case.

g-ortuno:
At least in the advertisement bytes there is a difference between an empty UUIDs list and a non existent UUIDs list. In the current spec both of these would result in the same event. Granted I don't think the distinction is that useful, I think we should try to keep the API as low-level as reasonable.

Empty UUID list:

| Value        | Notes           |
| ------------- |-------------|
| 0x02      | Length of this data |
| 0x01      | "Flags" Data Type Value |
| 0x01 | Flags Value (LE Limited Discoverable Flag set)     |
| 0x02 | Length of this data|
| 0x02 | "Incomplete UUID List" Data Type Value |
| 0x05 | Length of this data |
| 0x09 | "Name" Data Type Value|
| 0x65 | 'f' |
| 0x6F | 'o' |
| 0x6F | 'o' |

No UUID list:

| Value        | Notes           |
| ------------- |-------------|
| 0x02      | Length of this data |
| 0x01      | "Flags" Data Type Value |
| 0x01 | Flags Value (LE Limited Discoverable Flag set)     |
| 0x05 | Length of this data |
| 0x09 | "Name" Data Type Value|
| 0x65 | 'f' |
| 0x6F | 'o' |
| 0x6F | 'o' |

(That said, the BT spec says to interpret a missing UUID List as an empty incomplete list[1], so the UUIDs list member should be non-optional. The argument still applies to Service Data and Manufacturer Specific Data though.)

[1] Supplement to Bluetooth Core Specification | CSSv7, Part A 1.1 Service UUID

jyasskin:
I agree that if there are distinct ad packets, they should generally result in distinct `BluetoothAdvertisingEvent` objects. (Although we aren't doing that with incomplete vs complete UUID lists or shortened vs complete Local Names...)

For Manufacturer-specific Data, CSSv7 Part A.1.4 says each MSD item is 2 octets of Company Identifier Code followed by the manufacturer-specific data. To include multiple MSDs, you just send multiple items. So 0 MSDs seems like it's indistinguishable from a missing MSD list. And A.1.11 looks the same. Am I mis-reading this?

g-ortuno:
Oh I see. I agree that they are indistinguishable. I was thinking that it's possible to have an empty map but the spec requires the fields to be 2 or more octets long so having an empty map would go against the spec. For example the following adv packet would be invalid:

| Value        | Notes           |
| ------------- |-------------|
| 0x02      | Length of this data |
| 0x01      | "Flags" Data Type Value |
| 0x01 | Flags Value (LE Limited Discoverable Flag set)     |
| 0x02 | Length of this data|
| 0xFF | "Manufacturer Data" Data Type Value |
| 0x05 | Length of this data |
| 0x09 | "Name" Data Type Value|
| 0x65 | 'f' |
| 0x6F | 'o' |
| 0x6F | 'o' |

PR #423, opened 2018-12-28
Modify BluetoothLEScanOptions default values

dougt:
The default values of BluetoothLEScanOptions should be as if you didn't specify a filter to RequestLEScan().

### DIFF
--- a/scanning.bs
+++ b/scanning.bs
@@ -195,8 +195,8 @@ spec:web-bluetooth
   <pre class="idl">
     dictionary BluetoothLEScanOptions {
       sequence&lt;BluetoothLEScanFilterInit> filters;
-      boolean keepRepeatedDevices = false;
-      boolean acceptAllAdvertisements = false;
+      boolean keepRepeatedDevices = true;
+      boolean acceptAllAdvertisements = true;
     };
 
     partial interface Bluetooth {


reillyeon:
I believe the reasoning behind the current default values is to force developers to consider filtering in their initial implementation rather than defaulting the most permissive filter and allowing developers to add filtering later. See also [RequestDeviceOptions](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-requestdeviceoptions).

reillyeon:
There is an error thrown if neither `filters` nor `acceptAllDevices` is set so it doesn't particularly matter if the IDL allows `options` to be omitted because the steps will fail later. I think it's mostly a question of documentation: since the parameter is effectively required then it doesn't make sense for it to be marked optional.

scanning.bs
```diff
@@ -195,8 +195,8 @@ spec:web-bluetooth
   <pre class="idl">
     dictionary BluetoothLEScanOptions {
       sequence&lt;BluetoothLEScanFilterInit> filters;
-      boolean keepRepeatedDevices = false;
-      boolean acceptAllAdvertisements = false;
+      boolean keepRepeatedDevices = true;
```
jyasskin: https://heycam.github.io/webidl/#dfn-dictionary-member-default-value says, "As with operation argument default values, it is strongly suggested not to use true as the default value for boolean-typed dictionary members, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false)."

To change the default, we'd need to change the sense of the argument name.

As to whether it's a good idea, I arranged this to make the cheaper option take less typing.

jyasskin:
https://heycam.github.io/webidl/#dfn-optional-argument says,

> If the type of an argument is a dictionary type ..., and that dictionary type and its ancestors have no required members, and the argument is either the final argument or ..., then the argument must be specified as optional.

So, even though the argument isn't actually optional, the WebIDL compiler won't let us omit that keyword.

Issue #238, opened 2016-05-04
Missing API for GATT Write w/o Response


martijnthe:
There does not seem to be a way to specify that the GATT Write w/o Response operation should be used when calling `writeValue()`.

I don't think this should be "abstracted away" / left up to the UA implementation because this is the only GATT operation by which GATT clients can have more than one GATT packets in flight towards the GATT server. The other write operations have to be ack'd before the next write can be sent out, making it very slow.

Most applications that I've seen that need to optimize for throughput (Pebble smartwatch, Parrot's drones, ...), usually use GATT Write w/o Response in combination with GATT Value Notifications for data in the other direction.

Also related: https://github.com/WebBluetoothCG/web-bluetooth/issues/26

jyasskin:
Do you have devices that allow both Write with Response and With w/o Response on the same characteristic?

What behavior do you want when the requested write type isn't among the characteristic's properties?

I suspect the API shape will be `writeValue(value, {type: 'without-response'})`

martijnthe:
> Do you have devices that allow both Write with Response and Write w/o Response on the same characteristic?

Is possible and allowed by the BT spec.
I'm sure someone will come up with a good use case for having both enabled for the same characteristic.

> What behavior do you want when the requested write type isn't among the characteristic's properties?
I'd say if `with-response` was explicitly specified, error out the way you would do normally when an operation is attempted that isn't allowed. If it's not explicitly specified, fallback to write with response?

> I suspect the API shape will be writeValue(value, {type: 'without-response'})
Sounds good to me!

360fun:
Hi, I think I may got stuck in this issue: basically I'm trying to cache the GATT server connection, but when I write some data it executes the command and it disconnects right after! So the second time that I try to write a command using the cached variable, I get the error "Uncaught (in promise) DOMException: GATT Service no longer exists.".
The trick seems to re-connect every time the GATT server, but this adds latency and this don't allow applications like driving drones, cars, robots...since they need a real time feedback! :(

I started to guess why this happens and then I read that the kind of command in this case, for my device, is "without response", so, guessing, when the API doesn't get anything after the promise, it disconnects the GATT server. If this is right...is there a workaround or we need to wait an update in the API? :\

g-ortuno:
It would be great if you could open a Chromium issue with your problem. Here are the instructions:

https://www.chromium.org/developers/how-tos/file-web-bluetooth-bugs

[...]

PR #433, opened 2019-04-18
Add Characteristic Write w/o Response methods

beaufortfrancois:
Following https://github.com/WebBluetoothCG/web-bluetooth/issues/238, this PR adds two new methods to `BluetoothRemoteGATTCharacteristic` to write values with and without response.

Steps to get live preview:
1. Go to https://api.csswg.org/bikeshed/
2. In "Load from URL", enter https://raw.githubusercontent.com/WebBluetoothCG/web-bluetooth/write-response/index.bs
3. Click "Process" button

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3704,6 +3704,8 @@ spec: promises-guide-1
           getDescriptors(optional BluetoothDescriptorUUID descriptor);
         Promise&lt;DataView> readValue();
         Promise&lt;void> writeValue(BufferSource value);
+        Promise&lt;void> writeValueWithResponse(BufferSource value);
+        Promise&lt;void> writeValueWithoutResponse(BufferSource value);
         Promise&lt;BluetoothRemoteGATTCharacteristic> startNotifications();
         Promise&lt;BluetoothRemoteGATTCharacteristic> stopNotifications();
       };
@@ -3911,14 +3913,17 @@ spec: promises-guide-1
       </ol>
     </div>
 
-    <div algorithm="Characteristic.writeValue()">
+    <div algorithm>
       <p>
-        The <code><dfn method for="BluetoothRemoteGATTCharacteristic">writeValue(<var>value</var>)</dfn></code> method, when invoked,
-        MUST run the following steps:
+        To <dfn>WriteCharacteristicValue</dfn>(<span class="argument-list"><var>this</var>: BluetoothRemoteGATTCharacteristic,<br>
+        <var>value</var>: BufferSource,<br>
+        <var>withResponse</var>: boolean,<br>
+        <var>withoutResponse</var>: boolean),</span><br>
+        the UA MUST perform the following steps:
       </p>
       <ol>
         <li>
-          If <code>this.uuid</code> is <a>blocklisted for writes</a>,
+          If <code>|this|.uuid</code> is <a>blocklisted for writes</a>,
           return <a>a promise rejected with</a> a {{SecurityError}}
           and abort these steps.
         </li>
@@ -3933,13 +3938,13 @@ spec: promises-guide-1
           and abort these steps.
         </li>
         <li>
-          If <code>this.service.device.gatt.{{BluetoothRemoteGATTServer/connected}}</code> is `false`,
+          If <code>|this|.service.device.gatt.{{BluetoothRemoteGATTServer/connected}}</code> is `false`,
           return <a>a promise rejected with</a> a {{NetworkError}}
           and abort these steps.
         </li>
         <li>
           Let |characteristic| be
-          <code>this.{{[[representedCharacteristic]]}}</code>.
+          <code>|this|.{{[[representedCharacteristic]]}}</code>.
         </li>
         <li>
           If |characteristic| is `null`,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3997,6 +4024,45 @@ spec: promises-guide-1
...
+      </p>
+      <blockquote>
+        <a>WriteCharacteristicValue</a>(<span class="argument-list"><i>this</i>=<code>this</code>,<br>
+          <i>value</i>=<code><var>value</var></code>,<br>
+          <i>withResponse</i>=false,<br>
+          <i>withoutResponse</i>=false)</span>
+      </blockquote>
+    </div>
+
+    <div algorithm="Characteristic.writeValueWithResponse()">
```
jyasskin: These should probably be marked with `class="unstable"` until they pass an origin trial or Chrome ships them.

beaufortfrancois: Done

index.bs
```diff
@@ -3966,9 +3982,20 @@ spec: promises-guide-1
...
-              the <a>Characteristic Value Write</a> procedure
-              to write <var>bytes</var> to <var>characteristic</var>.
+              Write <var>bytes</var> to <var>characteristic</var> by performing
+              the following steps:
+              <ol>
+                <li>
+                  If |withResponse| is `true`, use the <a>Write Characteristic Value</a> procedure.
+                </li>
+                <li>
+                  Otherwise, if |withoutResponse| is `true`, use the <a>Write Without Response</a> procedure.
```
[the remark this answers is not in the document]
beaufortfrancois: Done.

index.bs
```diff
@@ -3911,14 +3913,17 @@ spec: promises-guide-1
...
 
-    <div algorithm="Characteristic.writeValue()">
+    <div algorithm>
       <p>
-        The <code><dfn method for="BluetoothRemoteGATTCharacteristic">writeValue(<var>value</var>)</dfn></code> method, when invoked,
-        MUST run the following steps:
+        To <dfn>WriteCharacteristicValue</dfn>(<span class="argument-list"><var>this</var>: BluetoothRemoteGATTCharacteristic,<br>
+        <var>value</var>: BufferSource,<br>
+        <var>withResponse</var>: boolean,<br>
+        <var>withoutResponse</var>: boolean),</span><br>
```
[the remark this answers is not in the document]
beaufortfrancois: Done.

beaufortfrancois:
@jyasskin Thanks for the review. I've addressed your feedback.

index.bs
```diff
@@ -3986,10 +3988,10 @@ spec: promises-guide-1
               the following steps:
               <ol>
                 <li>
-                  If |withResponse| is `true`, use the <a>Write Characteristic Value</a> procedure.
+                  If |response| is "required", use the <a>Write Characteristic Value</a> procedure.
```
jyasskin: You can use `<dl class="switch">` (https://w3c.github.io/tr-design/src/readme.html#switch) to make these branches a bit more readable.

index.bs
```diff
@@ -3911,14 +3913,16 @@ spec: promises-guide-1
       </ol>
     </div>
 
-    <div algorithm="Characteristic.writeValue()">
+    <div algorithm>
```
jyasskin: Removing the algorithm name will make some bikeshed errors harder to read.

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done.

index.bs
```diff
@@ -3933,28 +3937,42 @@ spec: promises-guide-1
...
+              abort these steps if any of the following is true:
+              <ol>
+                <li>
+                 |response| is "required" and <code>Write</code> bit is not set in <var>characteristic</var>'s <a lt="Characteristic Properties">properties</a>
+                </li>
+                <li>
+                 |response| is "never" and <code>Write Without Response</code> bit is not set in <var>characteristic</var>'s <a lt="Characteristic Properties">properties</a>
+                </li>
+                <li>
+                 |response| is "optional" and none of the <code>Write</code>, <code>Write Without Response</code> or <code>Authenticated Signed Writes</code> bits are set
```
reillyeon: The "Authenticated Signed Writes" bit appears equivalent to the "Write Without Response" bit so should be mentioned in the "never" section.

beaufortfrancois: You are absolutely right. Thanks for catching this!

beaufortfrancois:
@jyasskin @reillyeon Feedback addressed.

beaufortfrancois:
I'll be updating implementation status page as well: https://github.com/WebBluetoothCG/web-bluetooth/pull/436

index.bs
```diff
@@ -3966,9 +3985,23 @@ spec: promises-guide-1
...
+                <dd>
+                  Use the <a>Write Characteristic Value</a> procedure.
+                </dd>
+                <dt>If |response| is "never"</dt>
+                <dd>
+                  Use the <a>Write Without Response</a> procedure.
+                </dd>
+                <dt>Otherwise</dt>
+                <dd>
+                  Use any combination of the sub-procedures in
```
g-ortuno: Would be nice if we could specify which procedures exactly we should do here e.g.:
```
If the Write bit is set in characteristic’s properties, use the Write Characteristic Value procedure, otherwise use the Write without Response procedure.
```
I think currently we just let each platform decide, which creates differences across platforms and could result in differences across implementations.

g-ortuno: We could also implement `write()` in terms of `writeValueWithoutResponse()` and `writeValueWithResponse()`.

PR #477, opened 2020-02-21
Add logging and color to deploy script

odejesush:
This change makes errors in parsing the spec source files more visible
by adding more verbose logging with color.


jyasskin:
See https://github.com/w3cping/privacy-threat-model/blob/main/.travis.yml for a more recent way to deploy to gh-pages.

PR #478, opened 2020-03-09
Fix race condition in {,un}watchAdvertisements()

odejesush:
This change updates the algorithms for `watchAdvertisements()` and `unwatchAdvertisements()` to use the same parallel queue in order to prevent a race condition from happening when sending instructions to the OS to start/stop Bluetooth discovery for the device.

***

### :boom: Error: write EPROTO 139705468127104:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 12, 2020, 10:01 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWebBluetoothCG%2Fweb-bluetooth%2Fpull%2F478%2F695d9c2.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fodejesush%2Fweb-bluetooth%2Fpull%2F478.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WebBluetoothCG/web-bluetooth%23478.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2113,7 +2113,7 @@ following steps:
     1. [=Queue a task=] to perform the following steps:
         1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
         1. Resolve |promise| with `undefined`.
-1. [=If aborted=], reject |promise| with {{InvalidStateError}}.
+1. [=If aborted=], reject |promise| with {{AbortError}}.
 
 <div class="note">
 Scanning costs power, so websites should avoid watching for advertisements
--- a/index.bs
+++ b/index.bs
@@ -2090,7 +2090,7 @@ method, when invoked, MUST return <a>a new promise</a> |promise| and run the
 following steps <a>in parallel</a>:
 
 1. If <code>this.{{watchingAdvertisements}}</code> is `true`, resolve |promise|
-    with `undefined`.
+    with `undefined` and abort these steps.
 1. Ensure that the UA is scanning for this device's advertisements. The UA
     SHOULD NOT filter out "duplicate" advertisements for the same device.
 1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of the
@@ -2121,7 +2121,7 @@ as soon as possible.
 The <code><dfn method for="BluetoothDevice">unwatchAdvertisements()</dfn></code>
 method, when invoked, MUST run the following steps:
 
-1. If <code>this.{{watchingAdvertisements}}</code> is `false`, return.
+1. If <code>this.{{watchingAdvertisements}}</code> is `false`, abort these steps.
 1. Set <code>this.{{watchingAdvertisements}}</code> to `false`.
 1. If no more {{BluetoothDevice}}s in the whole UA have
     {{watchingAdvertisements}} set to `true`, the UA SHOULD stop scanning for
--- a/index.bs
+++ b/index.bs
@@ -2086,27 +2086,30 @@ attribute MUST perform the following steps:
 
 <div algorithm="watchAdvertisements" class="unstable">
 The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
-method, when invoked, MUST return <a>a new promise</a> |promise| and run the
-following steps <a>in parallel</a>:
+method, when invoked, MUST return <a>a new promise</a> |promise| and
+let |watchAdvertisementsQueue| be the result of
+<a href="https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue">
+starting a new parallel queue</a>:
 
-1. If <code>this.{{watchingAdvertisements}}</code> is `true`, resolve |promise|
-    with `undefined` and abort these steps.
-1. Ensure that the UA is scanning for this device's advertisements. The UA
-    SHOULD NOT filter out "duplicate" advertisements for the same device.
-1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of the
-    following errors, and abort these steps:
+1. <a href="https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task">
+    Enqueue the following steps</a> to |watchAdvertisementsQueue|:
+    1. If a <code>{{unwatchAdvertisements()}}</code> task is enqueued in
+        |watchAdvertisementsQueue| reject |promise| with {{InvalidStateError}}.
+    1. Ensure that the UA is scanning for this device's advertisements. The UA
+        SHOULD NOT filter out "duplicate" advertisements for the same device.
+    1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of
+        the following errors, and abort these steps:
 
-    <dl class="switch">
-      <dt>The UA doesn't support scanning for advertisements</dt>
-      <dd>{{NotSupportedError}}</dd>
+        <dl class="switch">
+          <dt>The UA doesn't support scanning for advertisements</dt>
+          <dd>{{NotSupportedError}}</dd>
 
-      <dt>Bluetooth is turned off</dt>
-      <dd>{{InvalidStateError}}</dd>
+          <dt>Bluetooth is turned off</dt>
+          <dd>{{InvalidStateError}}</dd>
 
-      <dt>Other reasons</dt>
-      <dd>{{UnknownError}}</dd>
-    </dl>
-1. <a>Queue a task</a> to perform the following steps:
+          <dt>Other reasons</dt>
+          <dd>{{UnknownError}}</dd>
+        </dl>
     1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
     1. Resolve |promise| with `undefined`.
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2089,6 +2089,8 @@ The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
 method, when invoked, MUST return <a>a new promise</a> |promise| and run the
 following steps <a>in parallel</a>:
 
+1. If <code>this.{{watchingAdvertisements}}</code> is `true`, resolve |promise|
+    with `undefined`.
```
reillyeon: "[...] and abort these steps."

index.bs
```diff
@@ -2119,6 +2121,7 @@ as soon as possible.
 The <code><dfn method for="BluetoothDevice">unwatchAdvertisements()</dfn></code>
 method, when invoked, MUST run the following steps:
 
+1. If <code>this.{{watchingAdvertisements}}</code> is `false`, return.
```
reillyeon: Rather than "return" the usual specification language is "and abort these steps."

odejesush:
That's true. I wasn't sure if this needed to be explicitly mentioned in the algorithms or not.

In that case, I will update the commit to prevent the race condition. Would the best way to do this be to change unwatchAdvertisements to return a promise as well?

jyasskin:
I think when I wrote this, I was trying to ensure that
1. Synchronously when `unwatch` is called, advertisement events stop being delivered to the device. That's achieved by https://webbluetoothcg.github.io/web-bluetooth/#advertising-event-algorithm:~:text=If%20deviceObj.watchingAdvertisements%20is%20false,%20abort%20these%20sub-steps.
2. When `watch` is called, advertisement events only start being delivered after the promise returns.

I suspect the right fix is say that the `watch` steps [abort when](https://infra.spec.whatwg.org/#algorithm-conditional-abort) `unwatch` is called on the same device. 

There's still a race between the instructions to the OS, which might be fixed by putting both in the same [parallel queue](https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue). The Web Bluetooth spec ought to be using those in lots of places to avoid internal races, but they post-date when I wrote the spec.

index.bs
odejesush: I removed the checks for `watchingAdvertisements` now. Instead this change updates the APIs to prevent a race condition from happening between these two APIs.

odejesush:
Please take another look.

index.bs
```diff
@@ -2086,25 +2086,30 @@ attribute MUST perform the following steps:
...
 The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
-method, when invoked, MUST return <a>a new promise</a> |promise| and run the
-following steps <a>in parallel</a>:
-
-1. Ensure that the UA is scanning for this device's advertisements. The UA
-    SHOULD NOT filter out "duplicate" advertisements for the same device.
-1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of the
-    following errors, and abort these steps:
+method, when invoked, MUST return <a>a new promise</a> |promise| and
+let |watchAdvertisementsQueue| be the result of
```
jyasskin: Imitate https://html.spec.whatwg.org/multipage/workers.html#shared-worker-manager to make a global parallel queue. What you have here would create a new one every time `watchAdvertisements()` is called.

index.bs
```diff
@@ -2086,25 +2086,30 @@ attribute MUST perform the following steps:
...
+<a href="https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue">
+starting a new parallel queue</a>:
+
+1. <a href="https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task">
+    Enqueue the following steps</a> to |watchAdvertisementsQueue|:
```
jyasskin:
```suggestion
1. [=parallel queue/Enqueue the following steps=] to |watchAdvertisementsQueue|:
```

index.bs
```diff
@@ -2086,25 +2086,30 @@ attribute MUST perform the following steps:
...
+    1. If a <code>{{unwatchAdvertisements()}}</code> task is enqueued in
+        |watchAdvertisementsQueue| reject |promise| with {{InvalidStateError}}.
```
jyasskin: I _think_ we don't want this check. If someone calls
```js
await device.watchAvertisements();
device.unwatchAdvertisements();
await device.watchAdvertisements();
```
I'd expect the device to be watching at the end of the sequence.

index.bs
```diff
@@ -2117,8 +2122,12 @@ as soon as possible.
...
 <div algorithm="unwatchAdvertisements" class="unstable">
 The <code><dfn method for="BluetoothDevice">unwatchAdvertisements()</dfn></code>
-method, when invoked, MUST run the following steps:
+method, when invoked, MUST let |watchAdvertisementsQueue| be the result of
+<a href="https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue">
+starting a new parallel queue</a>:
 
+1. <a href="https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task">
+    Enqueue the following steps</a> to |watchAdvertisementsQueue|:
 1. Set <code>this.{{watchingAdvertisements}}</code> to `false`.
```
jyasskin: I think this should happen synchronously, rather than inside the parallel queue.

index.bs
```diff
@@ -2086,25 +2086,30 @@ attribute MUST perform the following steps:
...
 
-      <dt>Bluetooth is turned off</dt>
-      <dd>{{InvalidStateError}}</dd>
+          <dt>Bluetooth is turned off</dt>
+          <dd>{{InvalidStateError}}</dd>
 
-      <dt>Other reasons</dt>
-      <dd>{{UnknownError}}</dd>
-    </dl>
-1. <a>Queue a task</a> to perform the following steps:
```
jyasskin: You still need to queue these steps, so they happen back on the main thread.

index.bs
```diff
@@ -2086,25 +2086,30 @@ attribute MUST perform the following steps:
...
-following steps <a>in parallel</a>:
-
-1. Ensure that the UA is scanning for this device's advertisements. The UA
-    SHOULD NOT filter out "duplicate" advertisements for the same device.
-1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of the
-    following errors, and abort these steps:
+method, when invoked, MUST return <a>a new promise</a> |promise| and
+let |watchAdvertisementsQueue| be the result of
+<a href="https://html.spec.whatwg.org/multipage/infrastructure.html#starting-a-new-parallel-queue">
+starting a new parallel queue</a>:
```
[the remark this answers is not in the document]
odejesush: Thanks, I tried `<a lt="starting a new parallel queue">` initially but that didn't compile properly. Now I know the proper syntax to use.

index.bs
odejesush: Does the parallel queue guarantee that `watchAdvertisements()` finishes before the next `{,un}watchAdvertisements()` call?

odejesush: I'm confused on how watch will be aborted when unwatch is called.

index.bs
odejesush: I see, thanks!

index.bs
odejesush: Did you mean just this step, or the `unwatchAdvertisements()` algorithm?

odejesush:
Thanks Jeffrey. Let me know if I understood your comments correctly.

index.bs
```diff
@@ -2084,29 +2084,34 @@ attribute MUST perform the following steps:
...
 </div>
 
+A user agent has an associated <dfn>watch advertisements manager</dfn> which is
+the result of [=starting a new parallel queue=].
+
 <div algorithm="watchAdvertisements" class="unstable">
 The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
 method, when invoked, MUST return <a>a new promise</a> |promise| and run the
-following steps <a>in parallel</a>:
+follwing steps:
```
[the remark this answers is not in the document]
odejesush: Whoops, thanks!

odejesush:
Thanks for you help Jeffrey!

index.bs
```diff
@@ -2112,6 +2113,7 @@ follwing steps:
     1. [=Queue a task=] to perform the following steps:
         1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
         1. Resolve |promise| with `undefined`.
+1. [=If aborted=], reject |promise| with {{InvalidStateError}}.
```
jyasskin: Consider [AbortError](https://heycam.github.io/webidl/#aborterror) here, but I don't feel strongly.

odejesush: AbortError makes more sense.

PR #480, opened 2020-03-16
Use AbortController for watchAdvertisements()

odejesush:
Refactor [`watchAdvertisements()`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-watchadvertisements) to take an [`AbortSignal`](https://dom.spec.whatwg.org/#abortsignal) parameter from an
[`AbortController`](https://dom.spec.whatwg.org/#abortcontroller)` to cancel the operation when the signal's aborted flag
is set. As a result, unwatchAdvertisements() is no longer needed as it's
operation is replaced by the AbortController [abort steps](https://dom.spec.whatwg.org/#abortsignal-add).

***
<a href="https://pr-preview.s3.amazonaws.com/odejesush/web-bluetooth/pull/480.html" title="Last updated on Mar 20, 2020, 5:03 PM UTC (6c16e07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/480/cbd2a81...odejesush:6c16e07.html" title="Last updated on Mar 20, 2020, 5:03 PM UTC (6c16e07)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2136,8 +2136,7 @@ run the following steps:
                   perform the following steps, and abort these steps:
                   1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
                       `'not-watching'`.
-                  1. [=Reject=] |promise| with one of the following errors and
-                      abort these steps:
+                  1. [=Reject=] |promise| with one of the following errors:
 
                       <dl class="switch">
                         : The UA doesn't support scanning for advertisements
--- a/index.bs
+++ b/index.bs
@@ -2034,6 +2034,19 @@ described in the following table:
       a URL on that origin.
     </td>
   </tr>
+  <tr>
+    <td><dfn>\[[watchAdvertisementsState]]</dfn></td>
+    <td>`'not-watching'`</td>
+    <td>
+      An string enumeration describing the current state of a
+      {{BluetoothDevice/watchAdvertisements()}} operation. The possible
+      enumeration values are:
+      * `'not-watching'`
+      * `'pending-watch'`
+      * `'watching'`
+
+    </td>
+  </tr>
 </table>
 
 <div algorithm="get or create BluetoothDevice">
@@ -2101,31 +2114,49 @@ run the following steps:
     the following sub-steps:
     1. If <code>|options|.{{AbortSignal|signal}}</code>'s [=AbortSignal/aborted
         flag=] is set, then [=abort watchAdvertisements=] with `this` and
-        |promise|, and abort these steps.
+        abort these steps.
     1. [=AbortSignal/add|Add the following abort steps=] to <code>
         |options|.{{AbortSignal|signal}}</code>:
         1. Set |locallyAborted| to `true`.
-        1. [=Abort watchAdvertisements=] with `this` and |promise|.
+        1. [=Abort watchAdvertisements=] with `this`.
+        1. [=Reject=] |promise| with {{AbortError}}
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
     manager=], but abort when |locallyAborted| becomes `true`:
-    1. Ensure that the UA is scanning for this device's advertisements. The UA
-        SHOULD NOT filter out "duplicate" advertisements for the same device.
-    1. If the UA fails to enable scanning, <a>reject</a> |promise| with one of
-        the following errors, and abort these steps:
+    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is
+        `'not-watching'`, perform the following steps:
+        1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+            `'pending-watch'`.
+        1. Ensure that the UA is scanning for this device's advertisements.
+            The UA SHOULD NOT filter out "duplicate" advertisements for the same
+            device.
+        1. If the UA fails to enable scanning, perform the following steps:
+            1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+                `'not-watching'`.
+            1. [=Reject=] |promise| with one of the following errors and
+                abort these steps:
 
-        <dl class="switch">
-          <dt>The UA doesn't support scanning for advertisements</dt>
-          <dd>{{NotSupportedError}}</dd>
+                <dl class="switch">
+                  <dt>The UA doesn't support scanning for advertisements</dt>
+                  <dd>{{NotSupportedError}}</dd>
 
-          <dt>Bluetooth is turned off</dt>
-          <dd>{{InvalidStateError}}</dd>
+                  <dt>Bluetooth is turned off</dt>
+                  <dd>{{InvalidStateError}}</dd>
 
-          <dt>Other reasons</dt>
-          <dd>{{UnknownError}}</dd>
-        </dl>
-    1. [=Queue a task=] to perform the following steps, but [=abort when=] |locallyAborted| becomes `true`:
-        1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
-        1. Resolve |promise| with `undefined`.
+                  <dt>Other reasons</dt>
+                  <dd>{{UnknownError}}</dd>
+                </dl>
+        1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+            `'watching'`.
+        1. [=Queue a task=] to perform the following steps, but [=abort when=]
+            |locallyAborted| becomes `true`:
+            1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
+            1. [=Resolve=] |promise| with `undefined`.
+    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is
+        `'pending-watch'`, [=reject=] |promise| with {{InvalidStateError}} and
+        abort these steps.
+    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is `'watching'`,
+        [=resolve=] |promise| with `undefined` and abort these steps.
+1. [=If aborted=], [=reject=] |promise| with {{AbortError}}.
 
 
 Note: Scanning costs power, so websites should avoid watching for advertisements
@@ -2135,15 +2166,13 @@ as soon as possible.
 
 <div algorithm="abort watchAdvertisements" class="unstable">
 To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> for a
-{{BluetoothDevice}} |device| and a {{Promise}} |promise|, run these steps:
+{{BluetoothDevice}} |device|, run these steps:
 
+1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+    `'not-watching'`.
 1. Set <code>|device|.{{watchingAdvertisements}}</code> to `false`.
 1. [=parallel queue/Enqueue the following steps=] to [=watch advertisements
     manager=]:
-    1. Reject |promise| with {{AbortError}}.
-
-        Note: This is a no-op if |promise| has already
-        fulfulled.
     1. If no more {{BluetoothDevice}}s in the whole UA have
         {{watchingAdvertisements}} set to `true`, the UA SHOULD stop scanning
         for advertisements. Otherwise, if no more {{BluetoothDevice}}s
--- a/index.bs
+++ b/index.bs
@@ -2154,10 +2154,9 @@ run the following steps:
                   1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
                   1. [=Resolve=] |promise| with `undefined`.
       : `'pending-watch'`
-      :: 1. [=Reject=] |promise| with {{InvalidStateError}} and abort these
-              steps.
+      :: 1. [=Reject=] |promise| with {{InvalidStateError}}.
       : `'watching'`
-      :: 1. [=Resolve=] |promise| with `undefined` and abort these steps.
+      :: 1. [=Resolve=] |promise| with `undefined`.
 
     </dl>
 1. [=If aborted=], [=reject=] |promise| with {{AbortError}}.
--- a/index.bs
+++ b/index.bs
@@ -2122,8 +2122,7 @@ run the following steps:
 
     <dl class="switch">
       : `'not-watching'`
-      :: 1. [=Queue a task=] to set <code>
-              this.{{[[watchAdvertisementsState]]}}</code> to
+      :: 1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
               `'pending-watch'`.
           1. [=parallel queue/Enqueue the following steps=] to the [=watch
               advertisements manager=], but abort when
@@ -2173,10 +2172,9 @@ as soon as possible.
 To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> for a
 {{BluetoothDevice}} |device|, run these steps:
 
-1. [=Queue a task=] to set perform the following steps:
-    1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
-        `'not-watching'`.
-    1. Set <code>|device|.{{watchingAdvertisements}}</code> to `false`.
+1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+    `'not-watching'`.
+1. Set <code>|device|.{{watchingAdvertisements}}</code> to `false`.
 1. [=parallel queue/Enqueue the following steps=] to [=watch advertisements
     manager=]:
     1. If no more {{BluetoothDevice}}s in the whole UA have
--- a/index.bs
+++ b/index.bs
@@ -114,7 +114,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/#
 spec: WebIDL; urlPrefix: https://heycam.github.io/webidl/#
     type: dfn
         text: a copy of the bytes held; url: dfn-get-buffer-source-copy
-        
+
 spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/browsing-the-web.html/#
     type: dfn
         text: initializing a new Document object; url: dfn-initialise-the-document-object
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2088,13 +2087,18 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
 the result of [=starting a new parallel queue=].
 
 <div algorithm="watchAdvertisements" class="unstable">
-The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
+The <code><dfn method for="BluetoothDevice">watchAdvertisements(|signal|)</dfn></code>
 method, when invoked, MUST return <a>a new promise</a> |promise| and run the
 following steps:
 
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
-    manager=], but [=abort when=] <code>this.{{unwatchAdvertisements()}}</code>
```
reillyeon: I'm not familiar with the patterns around parallel queues so I defer to @jyasskin here but it seems like we still wait to avoid beginning the steps to start watching advertisements while we are processing the steps to abort a previous request.

index.bs
```diff
@@ -1959,8 +1959,7 @@ or {{Bluetooth}} instance).
     readonly attribute DOMString? name;
     readonly attribute BluetoothRemoteGATTServer? gatt;
 
-    Promise<void> watchAdvertisements();
-    void unwatchAdvertisements();
+    Promise<void> watchAdvertisements(AbortSignal signal);
```
jyasskin: It's safer for future evolution to use a dictionary options argument here

```suggestion
    Promise<void> watchAdvertisements(optional WatchAdvertisementsOptions options = {});
```
```webidl
dictionary WatchAdvertisementsOptions{
  AbortSignal signal;
};
```

I lean against making the signal required.

index.bs
```diff
@@ -2122,13 +2126,16 @@ as soon as possible.
...
+<div algorithm="abort watchAdvertisements" class="unstable">
+To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> with a |promise|,
+run these steps:
 
 1. Set <code>this.{{watchingAdvertisements}}</code> to `false`.
 1. [=parallel queue/Enqueue the following step=] to [=watch advertisements
     manager=]:
+    1. Reject |promise| with {{AbortError}}.
+        <div class="note">Note: This is a no-op if |promise| has already
+        fulfulled</div>
```
jyasskin: This should [do the `class=note` by itself](https://tabatkins.github.io/bikeshed/#notes-etc) if you use:

```suggestion

        Note: This is a no-op if |promise| has already fulfulled
```
(note the blank line)

index.bs
```diff
@@ -2122,13 +2126,16 @@ as soon as possible.
...
 </div>
 
-<div algorithm="unwatchAdvertisements" class="unstable">
-The <code><dfn method for="BluetoothDevice">unwatchAdvertisements()</dfn></code>
-method, when invoked, MUST run the following steps:
+<div algorithm="abort watchAdvertisements" class="unstable">
+To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> with a |promise|,
+run these steps:
 
 1. Set <code>this.{{watchingAdvertisements}}</code> to `false`.
```
jyasskin: There's no longer a `this` with this algorithm; you'll have to pass the |device| explicitly.

index.bs
```diff
@@ -2088,13 +2087,18 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
 following steps:
 
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
-    manager=], but [=abort when=] <code>this.{{unwatchAdvertisements()}}</code>
-    is called:
+    manager=]:
+    1. If |signal| [=AbortSignal/aborted flag=] is set, then:
+        1. [=Abort watchAdvertisements=] with |promise| and abort these steps.
+    1. Let |locallyAborted| be `false`.
+    1. [=AbortSignal/add|Add the following abort steps=] to |signal|.
```
jyasskin: You should add the abort steps in the main thread, before enqueuing the steps on the parallel queue. I think the steps within the parallel queue should probably say "but abort when |locallyAborted| becomes `true`."

index.bs
```diff
@@ -2088,13 +2087,18 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
+The <code><dfn method for="BluetoothDevice">watchAdvertisements(|signal|)</dfn></code>
 method, when invoked, MUST return <a>a new promise</a> |promise| and run the
 following steps:
 
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
-    manager=], but [=abort when=] <code>this.{{unwatchAdvertisements()}}</code>
-    is called:
+    manager=]:
+    1. If |signal| [=AbortSignal/aborted flag=] is set, then:
+        1. [=Abort watchAdvertisements=] with |promise| and abort these steps.
```
jyasskin: I think with this step:

```js
await device.watchAdvertisements();
let controller = new AbortController();
controller.abort();
device.watchAdvertisements({signal: controller.signal});
```

will lead to the device not-watching advertisements. That's a little weird, but maybe the right thing to do? Otherwise, the second call should probably reject its promise because `device.watchingAdvertisements==true`?

index.bs
jyasskin: I'm not really an expert on the patterns around parallel queues either. I think they're not widely used yet, so we should just do what seems right, and maybe bug Domenic or Jake for review when we think it's reasonable.

index.bs
odejesush: I think that it would be the right thing to do. Otherwise, I don't see a way to cancel the operation if `watchAdvertisements()` is called without an `AbortSignal`. (Unless we make `AbortSignal` a required parameter).

index.bs
odejesush: If the signal is optional, how would `watchAdvertisements()` be cancelled?

jyasskin: If someone calls `watchAdvertisements(new AbortController().signal)` they can't cancel it either, and even if they keep the controller around, we can't force them to cancel it. At best requiring the signal is a reminder to think about stopping the scan, but I suspect there are apps that legitimately don't need to stop scanning.

lopezbeto:
Sent from Yahoo Mail on Android 
 
  On Tue, Mar 17, 2020 at 10:00 AM, Ovidio Ruiz-Henríquez<notifications@github.com> wrote:   


@odejesush commented on this pull request.

In index.bs:
>  method, when invoked, MUST return <a>a new promise</a> |promise| and run the
 following steps:
 
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
-    manager=], but [=abort when=] <code>this.{{unwatchAdvertisements()}}</code>
-    is called:
+    manager=]:
+    1. If |signal| how i can stop all thid[=AbortSignal/aborted flag=] is set, then:
+        1. [=Abort watchAdvertisements=] with |promise| and abort these steps.

I think that it would be the right thing to do. Otherwise, I don't see a way to cancel the operation if watchAdvertisements() is called without an AbortSignal. (Unless we make AbortSignal a required parameter).

index.bs
odejesush: Thanks for the clarification.

index.bs
odejesush: Good catch, thanks!

index.bs
odejesush: I've asked Jake to take a look.

index.bs
```diff
@@ -2088,13 +2092,22 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
-
+The <code><dfn method for="BluetoothDevice">watchAdvertisements(|options|)</dfn>
+</code> method, when invoked, MUST return <a>a new promise</a> |promise| and
+run the following steps:
+
+1. If <code>|options|.{{AbortSignal|signal}}</code> is present, then perform
+    the following sub-steps:
+    1. If <code>|options|.{{AbortSignal|signal}}</code>'s [=AbortSignal/aborted
+        flag=] is set, then [=abort watchAdvertisements=] with `this` and
+        |promise| and abort these steps.
```
reillyeon: Use a comma to separate the two lists: "with `this` and |promise|, and abort these steps."

index.bs
```diff
@@ -2113,22 +2126,24 @@ following steps:
     1. [=Queue a task=] to perform the following steps:
         1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
         1. Resolve |promise| with `undefined`.
-1. [=If aborted=], reject |promise| with {{AbortError}}.
 
-<div class="note">
-Scanning costs power, so websites should avoid watching for advertisements
+
+Note: Scanning costs power, so websites should avoid watching for advertisements
 unnecessarily, and should call {{unwatchAdvertisements()}} to stop using power
```
reillyeon: "[...]and should use their `AbortController` to stop[...]"

odejesush: Good catch, thanks!

index.bs
reillyeon: Comma before the "and".

lopezbeto:
Sent from Yahoo Mail on Android 
 
  On Tue, Mar 17, 2020 at 12:48 PM, Reilly Grant<notifications@github.com> wrote:   


@reillyeon commented on this pull request.

In index.bs:
> @@ -2088,13 +2092,22 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
 the result of [=starting a new parallel queue=].
 
 <div algorithm="watchAdvertisements" class="unstable">
-The <code><dfn method for="BluetoothDevice">watchAdvertisements()</dfn></code>
-method, when invoked, MUST return <a>a new promise</a> |promise| and run the
-following steps:please  stop on this
-
+The <code><dfn method for="BluetoothDevice">watchAdvertisements(|options|)</dfn>
+</code> method, when invoked, MUST return <a>a new promise</a> |promise| and
+run the following steps:
+
+1. If <code>|options|.{{AbortSignal|signal}}</code> is present, then perform
+    the following sub-steps:
+    1. If <code>|options|.{{AbortSignal|signal}}</code>'s [=AbortSignal/aborted
+        flag=] is set, then [=abort watchAdvertisements=] with `this` and
+        |promise| and abort these steps.

Comma before the "and".

jakearchibald:
The use of abort signal & parallel queue LGTM.

odejesush:
Thanks for taking a look Jake.

index.bs
```diff
@@ -2088,13 +2092,22 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
-
+The <code><dfn method for="BluetoothDevice">watchAdvertisements(|options|)</dfn>
+</code> method, when invoked, MUST return <a>a new promise</a> |promise| and
+run the following steps:
+
+1. If <code>|options|.{{AbortSignal|signal}}</code> is present, then perform
+    the following sub-steps:
+    1. If <code>|options|.{{AbortSignal|signal}}</code>'s [=AbortSignal/aborted
+        flag=] is set, then [=abort watchAdvertisements=] with `this` and
+        |promise| and, abort these steps.
```
reillyeon: "|promise|, and abort these steps."

The comma should go before the "and".

odejesush: Sorry about this, I thought that I had fixed this.

index.bs
```diff
@@ -2113,22 +2126,24 @@ following steps:
     1. [=Queue a task=] to perform the following steps:
```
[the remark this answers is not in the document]
jakearchibald: That's fair. I'd interpreted it as "it cannot be aborted once this task is queued"

index.bs
```diff
@@ -2088,13 +2092,22 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
+        |promise|, and abort these steps.
+    1. Let |locallyAborted| be `false`.
+    1. [=AbortSignal/add|Add the following abort steps=] to <code>
+        |options|.{{AbortSignal|signal}}</code>:
+        1. Set |locallyAborted| to `true`.
+        1. [=Abort watchAdvertisements=] with `this` and |promise|.
 1. [=parallel queue/Enqueue the following steps=] to the [=watch advertisements
-    manager=], but [=abort when=] <code>this.{{unwatchAdvertisements()}}</code>
-    is called:
+    manager=], but abort when |locallyAborted| becomes `true`:
```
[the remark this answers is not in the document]
odejesush: Here's my attempt at tracking the state of `watchAdvertisements`. It seemed to me that if `watchAdvertisements()` is called while the state is `enabling`, then it should reject with an error. `InvalidStateError` seemed the most logical to me. Similarly, if the state was already `watching`, then the promise should resolve. For aborting watch advertisements, no check was needed because we're going to set the state to `not watching` regardless of what state it's in.

index.bs
```diff
@@ -2088,46 +2105,73 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
-          <dd>{{InvalidStateError}}</dd>
+    manager=], but abort when |locallyAborted| becomes `true`:
+    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is
+        `'not-watching'`, perform the following steps:
+        1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+            `'pending-watch'`.
+        1. Ensure that the UA is scanning for this device's advertisements.
+            The UA SHOULD NOT filter out "duplicate" advertisements for the same
+            device.
+        1. If the UA fails to enable scanning, perform the following steps:
```
jyasskin: Change `this.[[watchAdvertisementsState]]` on the main thread (inside queued tasks) to avoid race conditions. Here, that implies:

```suggestion
        1. If the UA fails to enable scanning, [=queue a task=] to perform the following steps, and abort these steps:
```

And then remove the "abort these steps" from the "reject |promise|" step.

index.bs
```diff
@@ -2088,46 +2105,73 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
+                  <dt>Bluetooth is turned off</dt>
+                  <dd>{{InvalidStateError}}</dd>
+
+                  <dt>Other reasons</dt>
+                  <dd>{{UnknownError}}</dd>
+                </dl>
+        1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+            `'watching'`.
+        1. [=Queue a task=] to perform the following steps, but [=abort when=]
+            |locallyAborted| becomes `true`:
```
jyasskin: To keep changes to `this.[[watchAdvertisementsState]]` on the main thread, move this under the next item.

```suggestion
        1. [=Queue a task=] to perform the following steps, but [=abort when=]
            |locallyAborted| becomes `true`:
            1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
                `'watching'`.
```

index.bs
```diff
@@ -2088,46 +2105,73 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
-          <dt>The UA doesn't support scanning for advertisements</dt>
-          <dd>{{NotSupportedError}}</dd>
-
-          <dt>Bluetooth is turned off</dt>
-          <dd>{{InvalidStateError}}</dd>
+    manager=], but abort when |locallyAborted| becomes `true`:
+    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is
+        `'not-watching'`, perform the following steps:
+        1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+            `'pending-watch'`.
```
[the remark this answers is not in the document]
odejesush: Great suggestion!

odejesush:
> Sorry this review got so sprawling. I think this is basically right now.

No worries, thanks a lot for helping me clean up and improve these algorithms significantly.


In `watchAdvertisements()`, I moved the parallel queue step inside of the 'not-watching' state steps and after the state is set to 'pending-watch' because it made the abort condition make more sense. For example, if state was already `not-watching` then the abort condition is already met so it seemed a bit ambiguous as to what to do in that case. Let me know what you think of this change.

index.bs
```diff
@@ -2109,53 +2109,58 @@ The <code><dfn method for="BluetoothDevice">watchAdvertisements(|options|)</dfn>
...
-    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is
-        `'pending-watch'`, [=reject=] |promise| with {{InvalidStateError}} and
-        abort these steps.
-    1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is `'watching'`,
-        [=resolve=] |promise| with `undefined` and abort these steps.
+1. If <code>this.{{[[watchAdvertisementsState]]}}</code> is:
+
+    <dl class="switch">
+      : `'not-watching'`
+      :: 1. [=Queue a task=] to set <code>
```
jyasskin: Now that this isn't inside of a parallel queue anymore, it's running on the main thread and doesn't need a task. :)

index.bs
```diff
@@ -2168,9 +2173,10 @@ as soon as possible.
 To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> for a
 {{BluetoothDevice}} |device|, run these steps:
 
-1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
-    `'not-watching'`.
-1. Set <code>|device|.{{watchingAdvertisements}}</code> to `false`.
+1. [=Queue a task=] to set perform the following steps:
```
jyasskin: This overall algorithm always runs on the main thread, so no need to queue a task to modify these bits.

index.bs
odejesush: Ah I see. Thanks for the heads up!

index.bs
```diff
@@ -2088,46 +2105,77 @@ A user agent has an associated <dfn>watch advertisements manager</dfn> which is
...
+                      </dl>
+              1. [=Queue a task=] to perform the following steps, but [=abort
+                  when=] <code>this.{{[[watchAdvertisementsState]]}}</code>
+                  becomes `not-watching`:
+                  1. Set <code>this.{{[[watchAdvertisementsState]]}}</code> to
+                      `watching`.
+                  1. Set <code>this.{{watchingAdvertisements}}</code> to `true`.
+                  1. [=Resolve=] |promise| with `undefined`.
+      : `'pending-watch'`
+      :: 1. [=Reject=] |promise| with {{InvalidStateError}} and abort these
```
jyasskin: Since there are no subsequent steps, you can skip the "and abort these steps" for this and the next case.

odejesush: Haha, yes you're right. Thanks!

odejesush:
Thank you very much @jyasskin for the thorough review and guidance on this change. This algorithm looks so much better now.

PR #481, opened 2020-03-20
Abort Bluetooth discovery when focus is lost

odejesush:
This change adds a handler for loss of full activity that stops any on-going Bluetooth scans. This is to mitigate an identity risk that is present if two sites are watching for device advertisements and receive the same advertisement at the same time.

***
<a href="https://pr-preview.s3.amazonaws.com/odejesush/web-bluetooth/pull/481.html" title="Last updated on Mar 20, 2020, 5:36 PM UTC (7ce50e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/481/e8635ae...odejesush:7ce50e4.html" title="Last updated on Mar 20, 2020, 5:36 PM UTC (7ce50e4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2186,6 +2186,25 @@ To <dfn>abort {{BluetoothDevice/watchAdvertisements}}</dfn> for a
 </div>
 
 <div class="unstable">
+### Handling Document Loss of Full Activity ### {#handling-full-activity-loss}
+
+Operations that initiate a scan for Bluetooth devices may only run in a [=fully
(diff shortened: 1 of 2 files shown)


ariccio:
There's only one problem with this, and it's mildly annoying. If you happen to step into a javascript debugger while a pairing dialog is open, or when a PIN dialog is open, it cancels the operation. This can be quite bad for developer experience.

This also applies to native debugging, where it can be quite confusing. I was stepping through chromium in windbg - it managed to waste an hour or so of my time :)

PR #492, opened 2020-05-03
serviceUUID does not exist in definition of BluetoothLEScanFilter

motla:
I guess `serviceUUID` has been replaced by `services` (array of `UUID`) in `BluetoothLEScanFilter`. Please check this is ok before merging.

### DIFF
--- a/scanning.bs
+++ b/scanning.bs
@@ -754,8 +754,8 @@ settings object=] is no longer [=fully active=], it must run these steps:
           and <code>|filter|.namePrefix</code> is a prefix of it.
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


motla:
Maybe it would be wise to rename it `serviceUUIDs`?

odejesush:
Re: Renaming to `serviceUUIDs`, Web Bluetooth has already been implemented using `services` for the filter, so changing it would break existing uses of it. I think that `services` is a good name, since the values within in don't necessarily need to be UUIDs. They can also be aliases for a UUID. See the [ResolveUUIDName](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothuuid-getservice) algorithm in the Web Bluetooth spec for info on this.

PR #505, opened 2020-07-14
Filter Services and Manufacturer Data from AD

odejesush:
This change updates the specification so that Service UUID and Manufacturer data is filtered out from `BluetoothAdvertisingEvent`s depending on the `RequestDeviceOptions` used when the device permission was granted to the site.

This change also removes references to the Permissions API `request()` algorithm because it is no longer included in that spec and was causing Bikeshed errors.

***
<a href="https://pr-preview.s3.amazonaws.com/odejesush/web-bluetooth/pull/505.html" title="Last updated on Jul 30, 2020, 8:25 PM UTC (19e0431)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/505/373de24...odejesush:19e0431.html" title="Last updated on Jul 30, 2020, 8:25 PM UTC (19e0431)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -528,6 +528,7 @@ that UAs will choose not to prompt.
   dictionary RequestDeviceOptions {
     sequence<BluetoothLEScanFilterInit> filters;
     sequence<BluetoothServiceUUID> optionalServices = [];
+    sequence<unsigned short> optionalManufacturerData = [];
     boolean acceptAllDevices = false;
   };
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1682,38 +1669,6 @@ feature</a>'s permission-related algorithms and types are defined as follows:
     </xmp>
   </dd>
 
-  <dt><a>permission request algorithm</a></dt>
```
reillyeon, line 1685: This algorithm is still referenced by the scanning specification.

odejesush: Thanks, I forgot to update that spec.

odejesush:
The changes should be the same, I just rebased the commits to include the new Travis CI changes.

PR #523, opened 2020-09-19
Update scanning.bs : filter.services can be null

motla:
In https://webbluetoothcg.github.io/web-bluetooth/scanning.html#bluetoothlescanfilter-match
As `filter.services` can be null as per the `BluetoothLEScanFilter` definition, we should precise it like in the previous steps.

[Bikeshed preview](https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/WebBluetoothCG/web-bluetooth/e96df2f98be9a86c2e2d5578698c7de4e5f58cab/scanning.bs#bluetoothlescanfilter-match)

### DIFF
--- a/scanning.bs
+++ b/scanning.bs
@@ -727,7 +727,8 @@ settings object=] is no longer [=fully active=], it must run these steps:
           and <code>|filter|.namePrefix</code> is a prefix of it.
         </li>
         <li>
-          For each |uuid| in <code>|filter|.services</code>,
+          If <code>|filter|.services</code> is non-<code>null</code>,
+          for each |uuid| in <code>|filter|.services</code>,
           some <a>Service UUID</a> in |event| is equal to |uuid|.
         </li>
         <li>


scanning.bs
```diff
@@ -727,7 +727,8 @@ settings object=] is no longer [=fully active=], it must run these steps:
           and <code>|filter|.namePrefix</code> is a prefix of it.
         </li>
         <li>
-          For each |uuid| in <code>|filter|.services</code>,
+          If <code>|filter|.services</code> is non-<code>null</code>,
```
reillyeon, line 730: It is probably more correct to make `services` non-nullable, as `manufacturerData` and `serviceData` already are. That way this condition vacuously holds if the filter didn't specify any services because there are no UUIDs to check.

reillyeon (changes requested):

scanning.bs
motla: Ok but in this case we have to initialize `services` to an empty array if it's not defined by the user.
https://github.com/WebBluetoothCG/web-bluetooth/blob/acd6ca76694b3ba209af9bd932114afc1c889123/scanning.bs#L415-L417

`, or an empty {{FrozenArray}} otherwise.` ?

reillyeon: Sounds good to me.

motla:
Switching to new PR to avoid merge conflict

PR #541, opened 2021-03-30
Configure GitHub Actions

reillyeon:
This change configures GitHub Actions to build the specification and associated documents using Bikeshed through [Spec Prod](https://github.com/w3c/spec-prod).

### DIFF
--- a/.github/workflows/pr-push.yml
+++ b/.github/workflows/pr-push.yml
@@ -1,0 +1,35 @@
+name: CI
+on:
+  pull_request: {}
+  push:
+    branches: [main]
+jobs:
+  main:
+    name: Build, Validate and Deploy
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: w3c/spec-prod@v2
+        with:
+          SOURCE: index.bs
+          DESTINATION: index.html
+          TOOLCHAIN: bikeshed
+          GH_PAGES_BRANCH: gh-pages
+      - uses: w3c/spec-prod@v2
+        with:
+          SOURCE: scanning.bs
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request: {}
+  push:
+    branches: [main]
+jobs:
+  main:
+    name: Build, Validate and Deploy
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: w3c/spec-prod@v2
```
jyasskin, line 12: I suggest giving each `spec-prod` step a [name](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsname) so that it'll be easier to identify failures.

```suggestion
      - name: index.bs
        uses: w3c/spec-prod@v2
```

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          SOURCE: index.bs
+          DESTINATION: index.html
+          TOOLCHAIN: bikeshed
```
jyasskin, line 16: Pass `BUILD_FAIL_ON: warning` to match the current Travis failure option
https://github.com/w3c/spec-prod/blob/main/docs/options.md#build_fail_on

jyasskin (approved):
I assume you're planning to delete .travis.yml in a subsequent change?

You may need to iterate with real commits rather than PRs in order to get the gh-pages behavior right, although maybe `w3c/spec-prod` is debugged well enough that you won't.

PR #545, opened 2021-04-27
Use sequence instead of map for data filters

beaufortfrancois:
The more I'm playing with manufacturer data filters, the more I think we should go with a `sequence` instead of a `map` in [BluetoothLEScanFilterInit](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-bluetoothlescanfilterinit) for both `manufacturerData` and `serviceData`.

Because users can have different strings to express the same uint16 ('0' and '00' for instance), we'll have to check for the unicity of the key which is what we're doing already with a sequence.
Moreover, having a required `companyIdentifier` dictionary key seems more readable than a simple object key.

I've added some code samples below to give you an idea of what I'm experiencing.

```js
// It's not easy to understand what we're filtering on at first glance and the
// empty object isn't nice to read as well.
let manufacturerData = { 0x00e0: {} };
navigator.bluetooth.requestDevice({ filters: [{ manufacturerData }] });

// And with the base-10 serialization of a uint16, it's not much better.
// We would have to handle the `0x` case among others or let developers do it.
let manufacturerData = { "224": {} };
navigator.bluetooth.requestDevice({ filters: [{ manufacturerData }] });
```

And here's what I'm proposing:

```js
// [NEW] Simplest form.
let manufacturerData = [{ companyIdentifier: 0x00e0 }];
navigator.bluetooth.requestDevice({ filters: [{ manufacturerData }] });

// [NEW] Full filtering on manufacturer data.
let manufacturerData = [
  {
    companyIdentifier: 0x00e0,
    dataPrefix: new Uint8Array([0x01, 0x02, 0x03, 0x04]),
    mask: new Uint8Array([0xff, 0xff, 0xff, 0x00]),
  },
];
navigator.bluetooth.requestDevice({ filters: [{ manufacturerData }] });
```

@reillyeon What do you think?

***
<a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/pull/545.html" title="Last updated on Apr 28, 2021, 9:40 AM UTC (e7a6e1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/545/3758135...e7a6e1a.html" title="Last updated on Apr 28, 2021, 9:40 AM UTC (e7a6e1a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -519,14 +519,20 @@ that UAs will choose not to prompt.
     BufferSource mask;
   };
 
+  dictionary BluetoothManufacturerDataFilterInit : BluetoothDataFilterInit {
+    required unsigned short companyIdentifier;
+  };
+
+  dictionary BluetoothServiceDataFilterInit : BluetoothDataFilterInit {
+    required BluetoothServiceUUID service;
+  };
+
   dictionary BluetoothLEScanFilterInit {
     sequence<BluetoothServiceUUID> services;
     DOMString name;
     DOMString namePrefix;
-    // Maps unsigned shorts to BluetoothDataFilters.
-    object manufacturerData;
-    // Maps BluetoothServiceUUIDs to BluetoothDataFilters.
-    object serviceData;
+    sequence<BluetoothManufacturerDataFilterInit> manufacturerData;
+    sequence<BluetoothServiceDataFilterInit> serviceData;
   };
 
   dictionary RequestDeviceOptions {
@@ -585,11 +591,11 @@ device has to:
 * have a name starting with <dfn dict-member
     for="BluetoothLEScanFilterInit">namePrefix</dfn> if that member is present,
 * <div class="unstable"> advertise <a>manufacturer specific data</a> matching
-    all of the key/value pairs in <dfn dict-member
+    all of the values in <dfn dict-member
     for="BluetoothLEScanFilterInit">manufacturerData</dfn> if that member is
     present, and</div>
 * <div class="unstable"> advertise <a>service data</a> matching all of the
-    key/value pairs in <dfn dict-member
+    values in <dfn dict-member
     for="BluetoothLEScanFilterInit">serviceData</dfn> if that member is
     present.</div>
 
@@ -853,7 +859,7 @@ values of <var>filters</var> passed to
   <tr>
     <td>
       <pre highlight="js">
-        [{manufacturerData: {17: {}}}]
+[{ manufacturerData: [{ companyIdentifier: 17 }] }]
       </pre>
     </td>
     <td>D1</td>
@@ -861,7 +867,7 @@ values of <var>filters</var> passed to
   <tr>
     <td>
       <pre highlight="js">
-        [{serviceData: {"A": {}}}]
+[{ serviceData: [{ service: "A" }] }]
       </pre>
     </td>
     <td>D2</td>
@@ -869,8 +875,10 @@ values of <var>filters</var> passed to
   <tr>
     <td>
       <pre highlight="js">
-        [{manufacturerData: {17: {}}},
-          {serviceData: {"A": {}}}]
+[
+  { manufacturerData: [{ companyIdentifier: 17 }] },
+  { serviceData: [{ service: "A" }] },
+]
       </pre>
     </td>
     <td>D1, D2</td>
@@ -878,8 +886,12 @@ values of <var>filters</var> passed to
   <tr>
     <td>
       <pre highlight="js">
-        [{manufacturerData: {17: {}},
-          serviceData: {"A": {}}}]
+[
+  {
+    manufacturerData: [{ companyIdentifier: 17 }],
+    serviceData: [{ service: "A" }],
+  },
+]
       </pre>
     </td>
     <td><i>&lt;none></i></td>
@@ -887,9 +899,13 @@ values of <var>filters</var> passed to
   <tr>
     <td>
       <pre highlight="js">
-        [{manufacturerData: {
-            17: {dataPrefix: new Uint8Array([1, 2, 3])},
-        }}]
+[
+  {
+    manufacturerData: [
+      { companyIdentifier: 17, dataPrefix: new Uint8Array([1, 2, 3]) },
+    ],
+  },
+]
       </pre>
     </td>
     <td>D1</td>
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
Note that CI should succeed once https://github.com/WebBluetoothCG/web-bluetooth/pull/546 is merged.

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
...
-        |manufacturerId| is outside the range from 0–65535 inclusive, throw a
-        {{TypeError}} and abort these steps.
-    1. Let |dataFilter| be
-        <code>|filter|.manufacturerData[|key|]</code>, <a>converted
-        to an IDL value</a> of type {{BluetoothDataFilterInit}}. If this
-        conversion throws an exception, propagate it and abort these steps.
-    1. Let |canonicalizedDataFilter| be the result of <a
-        for="BluetoothDataFilterInit">canonicalizing</a> |dataFilter|,
+    present and 
+    <code>|filter|.{{BluetoothLEScanFilterInit/manufacturerData}}.length === 0</code>,
```
reillyeon, line 1435: A nit, as was pointed out to me on the most recent specifications I worked on, the correct syntax for referring to a member of a dictionary is `filter["manufacturerData"]` rather than `filter.manufacturerData`. We should do a pass to fix this over the whole specification but might as well avoid introducing new examples of the old style.

index.bs
```diff
@@ -853,85 +859,114 @@ values of <var>filters</var> passed to
...
   <tr>
     <td>
       <pre highlight="js">
-        [{manufacturerData: {
-            17: {dataPrefix: new Uint8Array([1, 2, 3])},
-        }}]
+[
+  {
+    manufacturerData: [
+      { companyIdentifier: 17, dataPrefix: new Uint8Array([1, 2, 3]) },
```
reillyeon, line 905: Put `companyIdentifier` and `dataPrefix` on separate lines in this and the next two examples to avoid this table column becoming too wide.

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
...
-        conversion throws an exception, propagate it and abort these steps.
-    1. Let |canonicalizedDataFilter| be the result of <a
-        for="BluetoothDataFilterInit">canonicalizing</a> |dataFilter|,
+    present and 
+    <code>|filter|.{{BluetoothLEScanFilterInit/manufacturerData}}.length === 0</code>,
+    throw a {{TypeError}} and abort these steps.
+1. For each |manufacturerData| in
+    <code>|filter|.{{BluetoothLEScanFilterInit/manufacturerData}}</code>, do the following sub-steps:
+    1. If <code>|manufacturerData|.{{BluetoothManufacturerDataFilterInit/companyIdentifier}}</code>
+        is present in |canonicalizedFilter|.manufacturerData,
```
reillyeon, line 1440: We should probably be more specific about what "present" means here. Something like, "If there exists a object |existing| where |existing|["companyIdentifier"] === |manufacturerData|["companyIdentifier"], throw..."

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
...
+        throw a {{TypeError}} and abort these steps.
+    1. Let |canonicalizedManufacturerDataFilter| be the result of <a
+        for="BluetoothDataFilterInit">canonicalizing</a> |manufacturerData|,
         <a>converted to an ECMAScript value</a>. If this throws an exception,
         propagate that exception and abort these steps.
     1. Call <a abstract-op>CreateDataProperty</a>
```
reillyeon, line 1446: Since `canonicalizedFilter["manufacturerData"]` is now a sequence rather than an object we can append to it.

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
         <code><var>filter</var>.namePrefix</code>.
 1. Set <code>|canonicalizedFilter|.manufacturerData</code> to `{}`.
```
reillyeon, line 1432: Initialize `canonicalizedFilter["manufacturerData"]` to an empty array rather than an empty object.

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
...
+        for="BluetoothDataFilterInit">canonicalizing</a> |manufacturerData|,
         <a>converted to an ECMAScript value</a>. If this throws an exception,
         propagate that exception and abort these steps.
     1. Call <a abstract-op>CreateDataProperty</a>
-        (|canonicalizedFilter|.manufacturerData, |key|,
-        |canonicalizedDataFilter|).
+        (|canonicalizedFilter|.manufacturerData,
+        <code>|manufacturerData|.{{BluetoothManufacturerDataFilterInit/companyIdentifier}}</code>,
+        |canonicalizedManufacturerDataFilter|).
 1. Set <code>|canonicalizedFilter|.serviceData</code> to `{}`.
```
reillyeon, line 1450: Same here.

index.bs
```diff
@@ -1400,51 +1431,42 @@ returned from the following steps:
...
-        If this conversion throws an exception, propagate it and abort these
-        steps.
-    1. Let |canonicalizedDataFilter| be the result of <a
-        for="BluetoothDataFilterInit">canonicalizing</a> |dataFilter|,
+    1. Let |canonicalizedServiceDataFilter| be the result of <a
+        for="BluetoothDataFilterInit">canonicalizing</a> |serviceData|,
         <a>converted to an ECMAScript value</a>. If this throws an exception,
         propagate that exception and abort these steps.
     1. Call <a
         abstract-op>CreateDataProperty</a>(|canonicalizedFilter|.serviceData,
```
reillyeon, line 1468: Same here.

scanning.bs
```diff
@@ -57,9 +57,9 @@ spec: webidl
 spec:web-bluetooth
     type: dfn
         text: read only arraybuffer
-    type:enum-value; for:PermissionName; text:"bluetooth"
 spec: permissions-1
     type: enum-value
+        text: "bluetooth"
```
reillyeon, line 62: In #546 you only removed the conflicting enum value above. Is this change also necessary?

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done

scanning.bs
beaufortfrancois: It's not anymore. Thanks for catching!

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done

index.bs
beaufortfrancois: Done

beaufortfrancois (commented):
Thanks for the review!

reillyeon (approved):

PR #548, opened 2021-05-07
Throw TypeError if dataPrefix is present and empty

beaufortfrancois:
As discussed with @reillyeon at https://chromium-review.googlesource.com/c/chromium/src/+/2830933/comment/29d40301_fc8f11ca, we may want to throw a TypeError when `dataPrefix` is present and empty to warn developers that the code they've written probably isn't doing what they think it is.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1481,6 +1481,9 @@ The result of <dfn for="BluetoothDataFilterInit" export>canonicalizing</dfn> the
 {{BluetoothDataFilterInit}} |filter|, is the {{BluetoothDataFilterInit}}
 returned from the following steps:
 
+1. If <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}</code> is present and
+    <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}.length === 0</code>,
+    throw a {{TypeError}} and abort these steps.
 1. If <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}</code> is present,
     let |dataPrefix| be <a>a copy of the bytes held</a> by
     <code>|filter|.dataPrefix</code>. Otherwise, let |dataPrefix| be an empty


index.bs
```diff
@@ -1481,6 +1481,9 @@ The result of <dfn for="BluetoothDataFilterInit" export>canonicalizing</dfn> the
 {{BluetoothDataFilterInit}} |filter|, is the {{BluetoothDataFilterInit}}
 returned from the following steps:
 
+1. If <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}</code> is present and
+    <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}.length === 0</code>,
```
reillyeon, line 1485: A `BufferSource` doesn't have a `length` attribute. Looking at https://heycam.github.io/webidl/#es-buffer-source-types it doesn't look like there is an existing algorithm to get the length of a buffer source so I would move this step to between when we use the "a copy of the bytes held" algorithm to create `dataPrefix` and when we set it to an empty sequence if the dictionary member is not present.

reillyeon (changes requested):

index.bs
beaufortfrancois: Fixed. Thanks for catching!

index.bs
```diff
@@ -1481,10 +1481,13 @@ The result of <dfn for="BluetoothDataFilterInit" export>canonicalizing</dfn> the
...
-1. If <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}</code> is present,
-    let |dataPrefix| be <a>a copy of the bytes held</a> by
-    <code>|filter|.dataPrefix</code>. Otherwise, let |dataPrefix| be an empty
-    sequence of bytes.
+1. If <code>|filter|.{{BluetoothDataFilterInit/dataPrefix}}</code> is not
+    present, let |dataPrefix| be an empty sequence of bytes. Otherwise,
+    do the following sub-steps:
+    1. Let |dataPrefix| be <a>a copy of the bytes held</a> by
+        <code>|filter|.dataPrefix</code>.
+    1. If <code>|dataPrefix|.length === 0</code>, throw a {{TypeError}} and
```
reillyeon, line 1489: |dataPrefix| is a sequence so I don't think using dot notation isn't appropriate here. You can just say "If the length of dataPrefix is zero..."

reillyeon (changes requested):

index.bs
beaufortfrancois: Done in https://github.com/WebBluetoothCG/web-bluetooth/pull/548/commits/1bbd541a5fdff3189cc915f8979182a68cabfb48

reillyeon (approved):

PR #554, opened 2021-07-20
Documented Chrome's on-demand pairing.

cmumford:
Secure characteristics may only be accessed on paired devices.
Note those platforms which currently pair on-demand when accessing
a secure characteristic.


reillyeon (approved):

reillyeon:
Marked as substantive again to test @cmumford's membership in the CG. It looks good now.

Issue #599, opened 2023-03-23
requestDevice filters to exclude devices.


scheib:
Application developers may wish to filter OUT devices they know are not the correct ones.  "I have an array with the names of some devices I don't want to get when I do navigator.bluetooth.requestDevice" - [requested on Stack Overflow](https://stackoverflow.com/questions/75815353).

It would be reasonable to see filters that can exclude, such as by adding an `excludeFilters` the request call:
```
navigator.bluetooth.requestDevice({
  filters: [{services: [A, B]}],
  excludeFilters: [
    {namePrefix: "Unwanted"},
    {name: "Unique Name"},
    {name: "Unique Name2"},
  ]
})
```

Which, given many devices discovered:
| Device | Device Name | Device Services |
|-----------|---------------------|----------------------- |
| D1 | \<none> | A |
| D2 | \<none> | A, B |
| D3 | "Unwanted thing" | A, B |
| D4 | "Unique Name2" | A, B |
| D5 | "Unique Name3" | A, B |

Would present only `D2` and `D5` to the user in the chooser to select.

beaufortfrancois:
I've started https://github.com/WebBluetoothCG/web-bluetooth/pull/600 to address this issue.

PR #600, opened 2023-03-24
Add exclusionFilters to RequestDeviceOptions

beaufortfrancois:
This PR is about a new `"exclusionFilters"` option in `navigator.bluetooth.requestDevice()` options so that developers can exclude from the browser picker some known devices that are known to not work as expected.

Note that the name comes from the WebHID spec: https://github.com/WICG/webhid/pull/93

FIX https://github.com/WebBluetoothCG/web-bluetooth/issues/599

***
<a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/pull/600.html" title="Last updated on Mar 24, 2023, 1:45 PM UTC (2677632)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/600/4cb5684...2677632.html" title="Last updated on Mar 24, 2023, 1:45 PM UTC (2677632)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -520,6 +520,7 @@ that UAs will choose not to prompt.
 
   dictionary RequestDeviceOptions {
     sequence<BluetoothLEScanFilterInit> filters;
+    sequence<BluetoothLEScanFilterInit> exclusionFilters;
     sequence<BluetoothServiceUUID> optionalServices = [];
     sequence<unsigned short> optionalManufacturerData = [];
     boolean acceptAllDevices = false;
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1244,6 +1334,9 @@ The <code><dfn method for="Bluetooth">requestDevice(<var>options</var>)
 </dfn></code> method, when invoked, MUST return <a>a new promise</a> |promise|
 and run the following steps <a>in parallel</a>:
 
+1. If <code>|options|.{{RequestDeviceOptions/exclusionFilters}}</code> is
+    present and <code>|options|.{{RequestDeviceOptions/filters}}</code> is not present,
+    <a>reject</a> |promise| with a {{TypeError}} and abort these steps.
 1. If <code>|options|.{{RequestDeviceOptions/filters}}</code> is present and
```
reillyeon, line 1340: Check `exclusionFilters` here as well, so it isn't allowed if `acceptAllDevices` is set.

reillyeon (changes requested):

index.bs
beaufortfrancois: I believe this is not necessary.
1. If `exclusionFilters` is present and `filters` is not, abort. 
2. It means now either `exclusionFilters` and `filters` are set, or `exclusionFilters` does not exist.
3. Continue existing...

Am I wrong?

reillyeon: Good point.

reillyeon (approved):

PR #611, opened 2023-08-21
Add manufacturer data blocklist

chengweih001:
This is a spec patch to use the manufacturer data block list in https://github.com/WebBluetoothCG/registries/blob/master/manufacturer_data_blocklist.txt.

The spec patch follows how we did for [crrev.com/c/3781118](https://crrev.com/c/3781118).

***
<a href="https://pr-preview.s3.amazonaws.com/chengweih001/web-bluetooth/pull/611.html" title="Last updated on Aug 28, 2023, 10:04 PM UTC (f6f52eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/611/7096b82...chengweih001:f6f52eb.html" title="Last updated on Aug 28, 2023, 10:04 PM UTC (f6f52eb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4722,7 +4722,7 @@ produced by the following algorithm:
         |result|[|companyIdentifier|] to be an empty list.
     1. Let |dataFilter| be the result of <a>parsing an advertising data filter</a> at
         |dataPrefixStr| if the result is not an error, otherwise return an error.
-    1. Append |dataFilter| |result|[|companyIdentifier|].
+    1. Append |dataFilter| to |result|[|companyIdentifier|].
 1. Return |result|.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1307,8 +1307,7 @@ A {{BluetoothDataFilterInit}} |filter1| is a
     1. If <code>|filter1|.{{BluetoothDataFilterInit/dataPrefix}}[|byteIndex|] &
         |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>
         it not equal to <code>|filter2|.{{BluetoothDataFilterInit/dataPrefix}}[|byteIndex|] &
-        |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>
-        , return `false`.
+        |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>, return `false`.
     1. Set |byteIndex| to <code>|byteIndex| + 1</code>.
 1. Return `true`.
 
@@ -4659,14 +4658,12 @@ The UA should re-fetch this file periodically, but it's unspecified how often.
 
 # Advertising Data Filter # {#advertising-data-filter}
 
-An advertising data filter is a string that represents a set of manufacturer or
-service data.
+An advertising data filter represents a way to match against manufacturer or service data.
 
 <div algorithm="parse an advertising data filter">
 To <dfn>parse an advertising data filter</dfn> from a <a>string</a> |input|, perform the following steps:
 
-1. Let |words| be the result of invoking {{String/split(separator, limit)}} on |input| with separator
-    `'/'`.
+1. Let |words| be |input| <a>strictly split</a> on `/`.
 1. If the length of |words| is not equal to `2`, return an error and
     abort these steps.
 1. If the length of |words|[0] is not equal to the length of |words|[1], return an error and
@@ -4674,18 +4671,20 @@ To <dfn>parse an advertising data filter</dfn> from a <a>string</a> |input|, per
 1. Let |prefixData| be |words|[0].
 1. Let |prefixMask| be |words|[1].
 1. If |prefixData| or |prefixMask| is not a sequence of <a>ascii lower hex digit</a>, return an error.
-1. Let |result| be a new {{BluetoothDataFilterInit}} dictionary.
-1. Set |result|[{{BluetoothDataFilterInit/dataPrefix}}] to an empty {{Uint8Array}}.
-1. Set |result|[{{BluetoothDataFilterInit/mask}}] to an empty {{Uint8Array}}.
 1. Let |prefixIndex| be `0`.
+1. let |dataList| be an empty list.
+1. let |maskList| be an empty list.
 1. While |prefixIndex| is less than the length of |prefixData|, do the following sub-steps:
     1. Let |data| be the result of interpreting the characters at index |prefixIndex| and <code>|prefixIndex|  + 1</code>
         of |prefixData| as a hexadecimal number.
     1. Let |mask| be the result of interpreting the characters at index |prefixIndex| and <code>|prefixIndex|  + 1</code>
         of |prefixMask| as a hexadecimal number.
-    1. Invoke |result|[{{BluetoothDataFilterInit/dataPrefix}}].{{Array/push()}} with |data|.
-    1. Invoke |result|[{{BluetoothDataFilterInit/dataPrefix}}].{{Array/push()}} with |mask|.
+    1. Append |data| to |dataList|.
+    1. Append |mask| to |maskList|.
     1. Set |prefixIndex| to `|prefixIndex| + 2`.
+1. Let |result| be a new {{BluetoothDataFilterInit}} dictionary.
+1. Set |result|[{{BluetoothDataFilterInit/dataPrefix}}] to an {{Uint8Array}} constructed with |dataList|.
+1. Set |result|[{{BluetoothDataFilterInit/mask}}] to an {{Uint8Array}} constructed with |maskList|.
 1. Return |result|.
 
 </div>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4633,15 +4657,88 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
...
-# The GATT Blocklist # {#the-gatt-blocklist}
+# Advertising Data Prefix # {#advertising-data-prefix}
+
+An advertising data prefix is a string that represents a set of manufacturer or
+service data.
+
+<div algorithm="converting an advertising data prefix string to a BluetoothDataFilterInit object">
+The result of <dfn>converting an advertising data prefix string to a BluetoothDataFilterInit object</dfn>
+at a string |prefix| is a {{BluetoothDataFilterInit}}, or an error produced by
+the following algorithm:
```
reillyeon, line 4668: I'm finding the word "prefix" here is confusing because I don't know what it is a prefix of. What do you think of this alternative wording:

> To **parse an advertising data filter** from a [string](https://infra.spec.whatwg.org/#string) *input*:

Note that I use the "string" concept from the [infra](https://infra.spec.whatwg.org/) spec rather than the ECMA262 string concept, since we are talking about strings in the abstract and not specifically Javascript strings. I also follow the style from that spec and don't explicitly define the type of the result value.

index.bs
```diff
@@ -4633,15 +4657,88 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
...
+    1. If |maskStr|[0] is not a lower case hexadecimal or |maskStr|[1] is not a lower case hexadecimal,
+        return an error.
+    1. Let |data| be the result of invoking {{Number}}.{{Number/parseInt(string, radix)}}
+        with |dataStr| and `16`, or return an error if a {{TypeError}} is thrown.
+    1. Let |mask| be the result of invoking {{Number}}.{{Number/parseInt(string, radix)}}
+        with |maskStr| and `16`, or return an error if a {{TypeError}} is thrown.
+    1. Invoke |dataArray|.{{Array/push()}} with |data|.
+    1. Invoke |maskArray|.{{Array/push()}} with |mask|.
+    1. Set |prefixIndex| to `|prefixIndex| + 2`.
+1. Return <code>{dataPrefix: new Uint8Array(|dataArray|), mask: new Uint8Array(|maskArray|)}</code>.
```
reillyeon, line 4697: Explicitly build the dictionary, like this:

1. Let *result* be a [new](https://webidl.spec.whatwg.org/#new)  [`BluetoothDataFilterInit`](https://webbluetoothcg.github.io/web-bluetooth/#dictdef-bluetoothdatafilterinit) dictionary.
2. Set *result*["[`dataPrefix`](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdatafilterinit-dataprefix)"] to...

index.bs
```diff
@@ -4633,15 +4657,88 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
...
+    1. If |dataStr|[0] is not a lower case hexadecimal or |dataStr|[1] is not a lower case hexadecimal,
+        return an error.
+    1. Let |maskStr| be |prefixMask|.{{String/substring()}} with
+        |prefixIndex| and `|prefixIndex|  + 2`.
+    1. If |maskStr|[0] is not a lower case hexadecimal or |maskStr|[1] is not a lower case hexadecimal,
+        return an error.
+    1. Let |data| be the result of invoking {{Number}}.{{Number/parseInt(string, radix)}}
+        with |dataStr| and `16`, or return an error if a {{TypeError}} is thrown.
+    1. Let |mask| be the result of invoking {{Number}}.{{Number/parseInt(string, radix)}}
+        with |maskStr| and `16`, or return an error if a {{TypeError}} is thrown.
```
reillyeon, line 4693: This can probably be simplified to using infra concepts like checking that *data* is a sequence of [ASCII hex digits](https://infra.spec.whatwg.org/#ascii-hex-digit). It looks like the HTML specification doesn't bother defining exactly how to parse hexadecimal numbers and for example in [Colors](https://html.spec.whatwg.org/#colours) just says to "interpret characters as a hexadecimal number". That's probably good enough.

index.bs
```diff
@@ -4633,15 +4657,88 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
...
-to restrict the set of GATT attributes a website can access.
+to restrict the set of GATT attributes and manufacturer data a website can access.
+
+A <dfn>valid company identifier string</dfn> is a group of lowercase hexadecimal digits that has at
+least one but no more than four hexadecimal digits. The official list of company identifies can be found on
+<a href="https://www.bluetooth.com/specifications/assigned-numbers/">Bluetooth Assigned Numbers website</a>.
+
+<div algorithm="parsing the manufacturer data blocklist">
+The result of <dfn>parsing the manufacturer data blocklist</dfn> at a URL <var>url</var> is a map
+from Company Identifier Code to an {{Array}} of {{BluetoothDataFilterInit}}, or an error,
```
reillyeon, line 4714: Similarly here, use Infra concepts like a [list](https://infra.spec.whatwg.org/#lists) rather than ECMA262 concepts like an Array.

reillyeon (changes requested):

index.bs

If we are using infra's string here, for the step of splitting the string |input| , should I avoid using `{{String/split(separator, limit)}}`? Instead write something like this?

```
Let |words| be |input| split on `/`.
```

index.bs

index.bs

index.bs

index.bs
reillyeon: You can use [strictly split](https://infra.spec.whatwg.org/#strictly-split).

index.bs
```diff
@@ -1293,6 +1293,27 @@ a {{BluetoothDataFilterInit}} |filter| if the following steps return `match`.
...
+1. Let |byteIndex| be `0`.
+1. While |byteIndex| is less than the length of |filter2|, do the following sub-steps:
+    1. If <code>|filter1|.{{BluetoothDataFilterInit/mask}}[|byteIndex|] &
+        |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>
+        is not equal to |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|], return `false`.
+    1. If <code>|filter1|.{{BluetoothDataFilterInit/dataPrefix}}[|byteIndex|] &
+        |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>
+        it not equal to <code>|filter2|.{{BluetoothDataFilterInit/dataPrefix}}[|byteIndex|] &
+        |filter2|.{{BluetoothDataFilterInit/mask}}[|byteIndex|]</code>
+        , return `false`.
```
reillyeon, line 1311: Move this comma to the end of the line above, as otherwise you end up with a space on the left side of the comma.

index.bs
```diff
@@ -4633,15 +4657,79 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
 https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_descriptors.txt</a>.
 The UA should re-fetch this file periodically, but it's unspecified how often.
 
-# The GATT Blocklist # {#the-gatt-blocklist}
+# Advertising Data Filter # {#advertising-data-filter}
+
+An advertising data filter is a string that represents a set of manufacturer or
```
reillyeon, line 4662: "represents a way to match against manufacturer or service data"

index.bs
```diff
@@ -4633,15 +4657,79 @@ href="https://github.com/WebBluetoothCG/registries/blob/master/gatt_assigned_des
...
+1. Set |result|[{{BluetoothDataFilterInit/dataPrefix}}] to an empty {{Uint8Array}}.
+1. Set |result|[{{BluetoothDataFilterInit/mask}}] to an empty {{Uint8Array}}.
+1. Let |prefixIndex| be `0`.
+1. While |prefixIndex| is less than the length of |prefixData|, do the following sub-steps:
+    1. Let |data| be the result of interpreting the characters at index |prefixIndex| and <code>|prefixIndex|  + 1</code>
+        of |prefixData| as a hexadecimal number.
+    1. Let |mask| be the result of interpreting the characters at index |prefixIndex| and <code>|prefixIndex|  + 1</code>
+        of |prefixMask| as a hexadecimal number.
+    1. Invoke |result|[{{BluetoothDataFilterInit/dataPrefix}}].{{Array/push()}} with |data|.
+    1. Invoke |result|[{{BluetoothDataFilterInit/dataPrefix}}].{{Array/push()}} with |mask|.
```
reillyeon, line 4687:
```suggestion
    1. Invoke |result|[{{BluetoothDataFilterInit/mask}}].{{Array/push()}} with |mask|.
```

reillyeon, line 4687: `push()` isn't defined for `Uint8Array`. Instead I would collect the prefix and mask into lists and then construct the final values from those lists.

index.bs
```diff
@@ -4686,6 +4778,36 @@ blocklist</a>'s value is an error, or the UUID maps to either
 "<code>exclude</code>" or "<code>exclude-writes</code>" in the <a>GATT
 blocklist</a>.
 
+<div algorithm="blocklisted manufacturer data">
+A manufacturer data |manufacturerData| is a <dfn>blocklisted manufacturer data</dfn> if the following steps
+return `blocked`:
+
+1. If the <a>Manufacturer Data blocklist</a>'s value is an error, return `false`.
```
reillyeon, line 4785:
```suggestion
1. If the <a>Manufacturer Data blocklist</a>'s value is an error, return `blocked`.
```

reillyeon (changes requested):

index.bs

index.bs

index.bs

index.bs

index.bs

reillyeon (approved):

PR #614, opened 2023-09-29
Explainer for requestLEScan for granted devices

chengweih001:
Hi Web Bluetooth owner,

Could you please review this explainer that we would like to expand the existing requestLEScan to support listing for advertisements from granted devices as an alternative to the exiting [WatchAdvertisements](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothdevice-watchadvertisements) API.

### DIFF
--- a/requestLEScan-for-granted-devices-explainer.md
+++ b/requestLEScan-for-granted-devices-explainer.md
@@ -4,7 +4,7 @@
 
 The Bluetooth [requestLEScan](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dom-bluetooth-requestlescan) API to allows web apps to listen for advertisements from any Bluetooth device. We propose an extension to this API that limits scanning to only devices that the web app has been granted permission to connect to. This is motivated by the following:
 
-* Using watchAdvertisements API to listen to advertisements from multiple granted devices requires writing redundant code, the API is using [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for stopping the scanning activity, which is not recommended, and the API is still in development trials, not yet available to all users.
+* Using the watchAdvertisements API to listen to advertisements from multiple granted devices, which we believe to be the most common scenario, requires complex code to start and manage multiple scanning sessions.
 * The new requestLEScan API for granted devices would allow developers to write clear and concise code for listening to advertisements from granted devices, and it would avoid the confusion of what API to use for listening advertisements from bluetooth devices when requestLEScan for surrounding devices is officially launched in the future.
 
 ## Interface
--- a/requestLEScan-for-granted-devices-explainer.md
+++ b/requestLEScan-for-granted-devices-explainer.md
@@ -1,0 +1,144 @@
+# Bluetooth requestLEScan API for granted devices
+
+## Motivation
+
+The Bluetooth requestLEScan API for granted devices is a proposed expansion of the existing [requestLEScan](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dom-bluetooth-requestlescan) API to allow web apps to listen for advertisements from Bluetooth devices that have been granted permission to be used by the app. This is motivated by the following:
+
+* Using watchAdvertisements API to listen to advertisements from multiple granted devices requires writing redundant code, the API is using [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for stopping the scanning activity, which is not recommended, and the API is still in development trials, not yet available to all users.
+* The new requestLEScan API for granted devices would allow developers to write clear and concise code for listening to advertisements from granted devices, and it would avoid the confusion of what API to use for listening advertisements from bluetooth devices when requestLEScan for surrounding devices is officially launched in the future.
+
+## Interface
+
+```diff
+dictionary BluetoothLEScanOptions {
+  sequence<BluetoothLEScanFilterInit> filters;
+  boolean keepRepeatedDevices = false;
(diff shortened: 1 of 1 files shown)


requestLEScan-for-granted-devices-explainer.md
```diff
@@ -0,0 +1,147 @@
+# Bluetooth requestLEScan API for granted devices
+
+## Motivation
+
+The Bluetooth requestLEScan API for granted devices is a proposed expansion of the existing [requestLEScan](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dom-bluetooth-requestlescan) API to allow web apps to listen for advertisements from Bluetooth devices that have been granted permission to be used by the app. This is motivated by the following:
```
reillyeon, line 5:
```suggestion
The Bluetooth [requestLEScan](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dom-bluetooth-requestlescan) API to allows web apps to listen for advertisements from any Bluetooth device. We propose an extension to this API that limits scanning to only devices that the web app has been granted permission to connect to. This is motivated by the following:
```

requestLEScan-for-granted-devices-explainer.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Using watchAdvertisements API to listen to advertisements from multiple granted devices requires writing redundant code, the API is using [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for stopping the scanning activity, which is not recommended, and the API is still in development trials, not yet available to all users.
```
reillyeon, line 7:
```suggestion
* Using the watchAdvertisements API to listen to advertisements from multiple granted devices, which we believe to be the most common scenario, requires complex code to start and manage multiple scanning sessions.
```

requestLEScan-for-granted-devices-explainer.md
```diff
@@ -0,0 +1,147 @@
...
+* The new requestLEScan API for granted devices would allow developers to write clear and concise code for listening to advertisements from granted devices, and it would avoid the confusion of what API to use for listening advertisements from bluetooth devices when requestLEScan for surrounding devices is officially launched in the future.
```
reillyeon, line 8:
```suggestion
* The requestLEScan API already supports scanning for advertisements from multiple devices and can be augmented with a way to filter advertisements to only devices the site has permission to connect to. This allows user agents to tailor the permission UI for the relevant scenario.
```

requestLEScan-for-granted-devices-explainer.md
```diff
@@ -0,0 +1,147 @@
...
+    sequence<BluetoothServiceUUID> services;
+    DOMString name;
+    DOMString namePrefix;
+    sequence<BluetoothManufacturerDataFilterInit> manufacturerData;
++   DOMString id;
+};
+
+```
+
+A new flag named **listenOnlyGrantedDevices** will be added to [BluetoothLEScanOptions](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dictdef-bluetoothlescanoptions) to indicate requestLEScan is being used for listening for granted devices when it is set to `true`. This flag is needed to avoid confusion with the current experimental requestLEScan API which is for surrounding devices in the environment (i.e. including both granted and ungranted devices). The default value for listenOnlyGrantedDevices is `false`.
```
reillyeon, line 30:
```suggestion
A new flag named **listenOnlyGrantedDevices** will be added to [BluetoothLEScanOptions](https://webbluetoothcg.github.io/web-bluetooth/scanning.html#dictdef-bluetoothlescanoptions) to indicate requestLEScan is being used for listening for granted devices when it is set to `true`. This flag allows the user agent to only prompt for user consent when the site is requesting information about devices it doesn't already have permission to access. The default value for listenOnlyGrantedDevices is `false`.
```

reillyeon (approved):
LGTM with a few suggestions for how this change is presented.

requestLEScan-for-granted-devices-explainer.md
```diff
@@ -0,0 +1,144 @@
...
+  boolean acceptAllAdvertisements = false;
++ boolean listenOnlyGrantedDevices = false;
+};
+
+dictionary BluetoothLEScanFilterInit {
+    sequence<BluetoothServiceUUID> services;
+    DOMString name;
+    DOMString namePrefix;
+    sequence<BluetoothManufacturerDataFilterInit> manufacturerData;
++   DOMString id;
```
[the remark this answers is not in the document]
chengweih001: Thanks for the suggestion! I will create another PR to have another section to cover alternative and why we decided not to.

In short, we thought about using `sequence<DOMString> ids` in the beginning of the proposal, however, we found it would make it harder if an app wants to do different filtering for different devices. For example, to filter like what we have in [this scenario of having co2 level sensor and temp sensor](https://github.com/WebBluetoothCG/web-bluetooth/blob/main/requestLEScan-for-granted-devices-explainer.md#:~:text=The%20app%20can,byte%20is%200x2.).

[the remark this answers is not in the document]
reillyeon: Jack is out so he's unlikely to get to the PR he mentioned above for a while but the reason for not adding a `sequence<BluetoothDevice>` or `sequence<DOMString>` to `BluetoothLEScanFilterInit` is that all of the properties in a filter are ANDed together (e.g. this is how services work) so specifying multiple devices in a single filter would never match anything.

The `listenOnlyGrantedDevices` gives developers the behavior they are most likely to want by default (all advertisements they are allowed to receive) while they can provide per-device filters if they want to.

Issue #616, opened 2023-11-22
Add requestDevice User Prompt BiDi Extension Section to Bluetooth Spec


outofambit:
## Summary
Specify a WebDriver BiDi extension that grants control over user prompts produced by the Web Bluetooth API.

## Location of spec prose
Precedent exists in webdriver classic for specifying extensions in the relevant specification of the given technology rather than in the webdriver spec. The BTT Working Group has signaled that this is how it should work for BiDi.

## Primary Goal
Allow a web developer to test their application using Web Bluetooth and automate responding to user prompts.

## How this is different from Permissions
This extension will automate responding to a user prompt produced by `navigator.bluetooth.requestDevice`, which allows a user to choose a specific device to connect to. A side effect of `requestDevice` is granting permission to access a specific device.

The WebDriver BiDi group is [already working on a general permissions extension](https://github.com/w3c/webdriver-bidi/issues/587). Automation of the browser powerful feature permissions required by `navigator.bluetooth.requestDevice` will be handled by that extension.

--------
<details><summary>References</summary>
<ul>
<li><a href='https://webbluetoothcg.github.io/web-bluetooth/'>https://webbluetoothcg.github.io/web-bluetooth/</a></li>
<li><a href='https://github.com/webbluetoothcg/web-bluetooth/'>https://github.com/webbluetoothcg/web-bluetooth/</a></li>
<li>WebBluetoothCG/web-bluetooth/issues/613</li>
<li>w3c/webdriver-bidi/pull/523</li>
<li>w3c/webdriver-bidi/issues/506</li>
<li><a href='https://w3c.github.io/webdriver-bidi/'>https://w3c.github.io/webdriver-bidi/</a></li>
<li><a href='https://w3c.github.io/webdriver-bidi/#command-browsingContext-handleUserPrompt'>https://w3c.github.io/webdriver-bidi/#command-browsingContext-handleUserPrompt</a></li>
<li><a href='https://w3c.github.io/webdriver-bidi/#event-browsingContext-userPromptOpened'>https://w3c.github.io/webdriver-bidi/#event-browsingContext-userPromptOpened</a></li>
<li><a href='https://www.w3.org/TR/webdriver2/#user-prompts'>https://www.w3.org/TR/webdriver2/#user-prompts</a></li>
[...]

Issue #621, opened 2024-03-12
Proposal does not reflect Chromium's dynamic device selection


jugglinmike-bocoup:
When Chromium prompts users to choose a device, the list of available devices may change while the prompt is active. The current wording of the proposal doesn't appear to support this behavior, because [it invokes the Permission API's "prompt the user to choose" with a static set of Bluetooth devices](https://github.com/WebBluetoothCG/web-bluetooth/blob/7b56ee87ffd57d22ae8d12ba4ddd7b94aeb996c5/index.bs#L1479-L1492).

reillyeon:
I've opened https://github.com/w3c/permissions/issues/447 because I think this needs to be resolved first in the Permissions specification.

PR #622, opened 2024-03-20
Add WebDriver BiDi extension

OrKoN:
Closes https://github.com/WebBluetoothCG/web-bluetooth/issues/616

***
<a href="https://pr-preview.s3.amazonaws.com/OrKoN/web-bluetooth/pull/622.html" title="Last updated on Apr 11, 2024, 5:07 AM UTC (2d8cfac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/622/7b56ee8...OrKoN:2d8cfac.html" title="Last updated on Apr 11, 2024, 5:07 AM UTC (2d8cfac)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,11 +111,6 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
         text: error; url: dfn-errors
         text: local end; url: dfn-local-ends
 
-<!-- WebDriver BiDi should be updated to export this definition. -->
-<!-- https://github.com/speced/bikeshed/issues/2802 -->
-spec: WEBDRIVER-BIDI; urlPrefix: https://w3c.github.io/webdriver-bidi/
-    type: dfn
-        text: set of sessions for which an event is enabled; url: set-of-sessions-for-which-an-event-is-enabled
 </pre>
 <pre class="link-defaults">
 spec:dom
--- a/index.bs
+++ b/index.bs
@@ -1488,8 +1488,8 @@ steps:
         <var>scanResult</var> if they <a>match a filter</a> in
         <var>uuidExclusionFilters</var>.
 1. Let |browsing context| be [=this=]'s [=relevant global object=]'s [=browsing context=].
-1. Let |prompt id| be a new unique opaque string.
-1. [=Trigger a prompt open event=] given |browsing context|, |prompt id|, and |scanResult|.
+1. Let |promptId| be a new unique opaque string.
+1. [=Trigger a prompt open event=] given |browsing context|, |promptId|, and |scanResult|.
 1. <p id="requestDevice-prompt">Even if <var>scanResult</var> is empty,
     <a>prompt the user to choose</a> one of the devices in |scanResult|,
     associated with |descriptor|, and let |device| be the result.
@@ -1504,7 +1504,7 @@ steps:
       to indicate interest and then perform a privacy-disabled scan to retrieve
       the name.
     </div>
-1. [=Trigger a prompt close event=] given |browsing context| and |prompt id|.
+1. [=Trigger a prompt close event=] given |browsing context| and |promptId|.
 1. The UA MAY <a>add |device| to |storage|</a>.
 
     <div class="note">
@@ -4846,14 +4846,14 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
-Issue: cddl snippetes use the "text" type instead of
+Issue: CDDL snippetes use the "text" type instead of
 "browsingContext.BrowsingContext" to allow indepedent programmatic
 processing of CDDL snippets. Currently, other modules cannot be
 referenced.
 
 ## The bluetooth module ## {#bluetooth-module}
 
-The bluetooth module contains commands for managing the remote end Bluetooth behavior.
+The <dfn>bluetooth</dfn> module contains commands for managing the remote end Bluetooth behavior.
 
 ### Types ### {#bidi-types}
 
@@ -4879,7 +4879,7 @@ bluetooth.RequestDeviceInfo = {
 The bluetooth.RequestDeviceInfo type represents a single device in a request device prompt.
 
 <div algorithm="serialize device">
-To <dfn>serialize device</dfn> given [=Bluetooth device=] |device|:
+To <dfn>serialize a device</dfn> given a {{BluetoothDevice}} |device|:
 
 1. Let |id| be the id of |device|.
 1. Let |name| be the name of |device|.
@@ -4895,15 +4895,15 @@ bluetooth.RequestDevicePrompt = text
 
 The bluetooth.RequestDevicePrompt is an identifier for a single prompt.
 
-A remote end has a <dfn>map of browsing context to device prompts</dfn> which is a [=/map=] whose keys are [=browsing context ids=] and values are [=device prompts=].
+A remote end has a <dfn>map of browsing contexts to device prompts</dfn> which is a [=/map=] whose keys are [=browsing context ids=] and values are [=device prompts=].
 
 <div algorithm="get a prompt">
-To <dfn>get a prompt</dfn> given |browsing context id| and |prompt id|:
+To <dfn>get a prompt</dfn> given |browsingContextId| and |promptId|:
 
-1. If [=map of browsing context to device prompts=][|browsing context id|] does not [=map/exist=]:
+1. If [=map of browsing contexts to device prompts=][|browsingContextId|] does not [=map/exist=]:
     1. Return [=error=] with [=error code=] [=no such prompt=].
-1. Let |prompt| be [=map of browsing context to device prompts=][|browsing context id|].
-1. If |prompt|'s [=device prompt id=] is not |prompt id|:
+1. Let |prompt| be [=map of browsing contexts to device prompts=][|browsingContextId|].
+1. If |prompt|'s [=device prompt id=] is not |promptId|:
     1. Return [=error=] with [=error code=] [=no such prompt=].
 1. Return [=success=] with data |prompt|.
 
@@ -4924,7 +4924,7 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
 
 1. Let |devices| be an empty [=/list=].
 1. For each |device| in |prompt|'s [=set of devices=].
-    1. Append the result of [=serialize device=] with |device|.
+    1. [=list/Append=] the result of [=serialize a device|serializing=] |device| to |devices|.
 1. Return |devices|.
 
 </div>
@@ -4936,10 +4936,10 @@ This specification extends the set of [=error codes=] from
 
 <dl>
   <dt><dfn>no such device</dfn>
-  <dd>Tried to reference an unknown [=Bluetooth device=].
+  <dd>Tried to reference an unknown {{BluetoothDevice}}.
 
   <dt><dfn>no such prompt</dfn>
-  <dd>Tried to reference an unknown prompt.
+  <dd>Tried to reference an unknown [=device prompt=].
 </dl>
 
 ### Commands ### {#bidi-commands}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4827,6 +4842,225 @@ attribute on the {{Navigator}} object may be used.
...
+
+## The bluetooth module ## {#bluetooth-module}
+
+The bluetooth module contains commands for managing the remote end Bluetooth behavior.
+
+### Types ### {#bidi-types}
+
+#### The bluetooth.RequestDevice Type #### {#bluetooth-requestdevice-type}
+
+```cddl
```
OrKoN, line 4857: TODO: turn code blocks into pre with cddl and local/remote cddl classes so that WebDriver BiDi tooling can pick it up

index.bs
```diff
@@ -4827,6 +4842,225 @@ attribute on the {{Navigator}} object may be used.
...
+   name: text,
+}
+```
+
+The bluetooth.RequestDeviceInfo type represents a single device in a request device prompt.
+
+<div algorithm="serialize device">
+To <dfn>serialize device</dfn> given [=Bluetooth device=] |device|:
+
+1. Let |id| be the id of |device|.
```
OrKoN, line 4879: What is exactly the name/ID of a device? Probably this should link to some definitions from [=Bluetooth device=]

index.bs
```diff
@@ -4827,6 +4842,200 @@ attribute on the {{Navigator}} object may be used.
...
+   device: bluetooth.RequestDevice,
+)
+
+bluetooth.HandleRequestDevicePromptCancelParameters = (
+   accept: false,
+)
+</pre>
+
+<div algorithm="remote end steps for bluetooth.handleRequestDevicePrompt">
+The [=remote end steps=] with |command parameters| are:
```
nondebug, line 4975: Let's use indexing syntax since [=command parameters=] is a map:

The [=remote end steps=] with [=command parameters=] |params| are:

1. Let |contextId| be |params|[\`"context"\`].
1. Let |promptId| be |params|[\`"prompt"\`].

index.bs
```diff
@@ -4827,6 +4842,200 @@ attribute on the {{Navigator}} object may be used.
...
+1. Let |context id| be the value of the <code>context</code> field of |command parameters|.
+1. Let |prompt id| be the value of the <code>prompt</code> field of |command parameters|.
+1. Let |prompt| be the result of [=trying=] to [=get a prompt=] with |context id| and |prompt id|.
+1. Let |accept| be the value of the <code>accept</code> field of |command parameters|.
+1. If |accept| is true:
+    1. Let |device id| be the value of the <code>device</code> field of |command parameters|.
+    1. Let |device| be the result of [=trying=] to [=match a device in prompt=] given |prompt| and |device id|.
+    1. Acknowledge |prompt| with |device|.
+1. Otherwise:
+    1. Dismiss |prompt|.
```
nondebug, line 4986: I think this algorithm should remove |prompt| from [=map of browsing context to device prompts=] once the prompt is acknowledged or dismissed.

index.bs
```diff
@@ -4827,6 +4842,200 @@ attribute on the {{Navigator}} object may be used.
...
+</pre>
+
+The bluetooth.RequestDeviceInfo type represents a single device in a request device prompt.
+
+<div algorithm="serialize device">
+To <dfn>serialize device</dfn> given [=Bluetooth device=] |device|:
+
+1. Let |id| be the id of |device|.
+1. Let |name| be the name of |device|.
+1. Return a [=map=] matching the <code>bluetooth.RequestDeviceInfo</code> production, with the <code>id</code> field set to |id|, and the <code>name</code> field set to |name|.
```
nondebug, line 4886: nit: [=maps=] are an ordered sequence of entries and don't have fields. How about:

1. Return a [=map=] matching the \<code>bluetooth.RequestDeviceInfo\</code> production, with \`"id"\` set to |id| and \`"name"\` set to |name|.

nondebug (changes requested):

index.bs
```diff
@@ -4827,6 +4842,200 @@ attribute on the {{Navigator}} object may be used.
...
+
+#### The bluetooth.RequestDevice Type #### {#bluetooth-requestdevice-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.RequestDevice = text
+</pre>
+
+The bluetooth.RequestDevice is an identifier for a single device in a request device prompt.
+
+A <dfn>device prompt</dfn> is a [=tuple=] consisting of a <dfn>device prompt id</dfn> (a string) and a <dfn>set of devices</dfn> (a [=/set=]). It represents a prompt which allows a user to [=prompt the user to choose|choose=] a [=Bluetooth device=].
```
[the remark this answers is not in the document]
OrKoN: Made RequestDeviceInfo.name nullable. Updated phrasing.

OrKoN:
thanks for reviewing @nondebug addressed the comments, PTAL

index.bs
```diff
@@ -4827,6 +4842,200 @@ attribute on the {{Navigator}} object may be used.
...
+For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
+
+Issue: cddl snippetes use the "text" type instead of
+"browsingContext.BrowsingContext" to allow indepedent programmatic
+processing of CDDL snippets. Currently, other modules cannot be
+referenced.
+
+## The bluetooth module ## {#bluetooth-module}
+
+The bluetooth module contains commands for managing the remote end Bluetooth behavior.
```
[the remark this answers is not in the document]
OrKoN: I have removed the definition because it conflicts with another definition with the same id. We also usually do not need the definition for the module, is there a place you want to reference it?

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+
+### Types ### {#bidi-types}
+
+#### The bluetooth.RequestDevice Type #### {#bluetooth-requestdevice-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.RequestDevice = text
+</pre>
+
+The bluetooth.RequestDevice is an identifier for a single device in a request device prompt.
```
reillyeon, line 4861:
```suggestion
A bluetooth.RequestDevice is an identifier for a single device in a request device prompt.
```

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+
+</div>
+
+#### The bluetooth.RequestDevicePrompt Type #### {#bluetooth-requestdeviceprompt-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.RequestDevicePrompt = text
+</pre>
+
+The bluetooth.RequestDevicePrompt is an identifier for a single prompt.
```
reillyeon, line 4895:
```suggestion
A bluetooth.RequestDevicePrompt is an identifier for a single prompt.
```

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+#### The bluetooth.RequestDeviceInfo Type #### {#bluetooth-requestdeviceinfo-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.RequestDeviceInfo = {
+   id: bluetooth.RequestDevice,
+   name: text / null,
+}
+</pre>
+
+The bluetooth.RequestDeviceInfo type represents a single device in a request device prompt.
```
reillyeon, line 4878:
```suggestion
A bluetooth.RequestDeviceInfo represents a single device in a request device prompt.
```

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+
+The bluetooth.RequestDevicePrompt is an identifier for a single prompt.
+
+A remote end has a <dfn>map of browsing contexts to device prompts</dfn> which is a [=/map=] whose keys are [=browsing context ids=] and values are [=device prompts=].
+
+<div algorithm="get a prompt">
+To <dfn>get a prompt</dfn> given |browsingContextId| and |promptId|:
+
+1. Let |promptMap| be the [=map of browsing contexts to device prompts=].
+1. If |promptMap|[|browsingContextId|] does not [=map/exist=]::
```
reillyeon, line 4903:
```suggestion
1. If |promptMap|[|browsingContextId|] does not [=map/exist=]:
```

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+
+bluetooth.RequestDevicePromptOpenedParameters = {
+   context: text,
+   prompt: bluetooth.RequestDevicePrompt,
+   devices: [* bluetooth.RequestDeviceInfo],
+}
+</pre>
+
+<div algorithm="remote end event trigger for bluetooth.requestDevicePromptOpened">
+To <dfn>trigger a prompt open event</dfn> given a [=browsing context=] |context|, a string |promptId|, and a [=/set=] of [=Bluetooth devices=] |devices|:
```
reillyeon, line 5025: This is unfortunately unimplementable due to the mismatch between the specification algorithm for "request Bluetooth devices" and how implementations actually have to implement a Bluetooth device selection prompt. As mentioned in #621, the set of Bluetooth devices in the prompt implemented by Chromium is dynamic. This is necessary because Bluetooth device discovery is a gradual process. It is not possible to instantaneously discover all of the nearby available Bluetooth devices.

At best, this could be implemented by waiting for the Bluetooth scan triggered by the appearance of the prompt to time out (usually 30 seconds) before triggering the `requestDevicePromptOpened` event. This would make automated tests quite slow.

The better solution is to introduce an interface similar to the [Chrome DevTools Protocol](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/device_access_handler.cc;l=94;drc=d7d69375c25df2dc3980e6a4edc5d032ec940efc) which generates a new event every time the set of available devices changes. Technically this doesn't require any changes to these steps but requires a note that this event can be fired multiple times for the same prompt ID.

reillyeon (changes requested):

index.bs
OrKoN: The implementation of this interface will be based on Chrome DevTools Protocol implementation so it will be also emitted multiple times with the updated list of the devices if CDP emits the underlying event multiple times. I could add a note and link #621 because it sounds like the main spec algorithms need to be updated to reflect device list updates. Alternatively, we could add an event called `requestDevicePromptUpdated` to handle updates separately (although I am not sure if the clients really have a use case to distinguish it). Sounds like specifying that event would be completely blocked on #621

index.bs
```diff
@@ -4827,6 +4837,205 @@ attribute on the {{Navigator}} object may be used.
...
+   params: bluetooth.RequestDevicePromptOpenedParameters 
+)
+
+bluetooth.RequestDevicePromptOpenedParameters = {
+   context: text,
+   prompt: bluetooth.RequestDevicePrompt,
+   devices: [* bluetooth.RequestDeviceInfo],
+}
+</pre>
+
```
OrKoN, line 5023:
```suggestion

Issue: In practice, the device list is dynamically updated while a prompt is open. The spec text currently does not reflect that but this event might be emitted multiple times with the same `prompt` ID and the fresh device list. See https://github.com/WebBluetoothCG/web-bluetooth/issues/621
```

OrKoN: @reillyeon Have I described it more or less accurately?

index.bs
reillyeon: I don't think we need a separate event but we should name it so that developers won't be surprised when it is called more than once. `requestDevicePromptUpdated` seems reasonable with the note you drafted above. Libraries can wrap it in friendler language-specific APIs similar to what Puppeteer has done: https://pptr.dev/api/puppeteer.devicerequestprompt

index.bs
```diff
@@ -1476,6 +1482,9 @@ steps:
     1. If |exclusionFilters| isn't `null`, remove devices from
         <var>scanResult</var> if they <a>match a filter</a> in
         <var>uuidExclusionFilters</var>.
+1. Let |browsingContext| be [=this=]'s [=relevant global object=]'s [=browsing context=].
+1. Let |promptId| be a new unique opaque string.
+1. [=Trigger a prompt updated event=] given |browsingContext|, |promptId|, and |scanResult|.
```
reillyeon, line 1487: Please add your proposed note about multiple update events here.

OrKoN: Done, added the note here.

index.bs
```diff
@@ -1476,6 +1482,16 @@ steps:
     1. If |exclusionFilters| isn't `null`, remove devices from
         <var>scanResult</var> if they <a>match a filter</a> in
         <var>uuidExclusionFilters</var>.
+1. Let |browsingContext| be [=this=]'s [=relevant global object=]'s [=browsing context=].
+1. Let |promptId| be a new unique opaque string.
+
+Issue: In practice, the device list is dynamically updated while a
+prompt is open. The spec text currently does not reflect that but this
+event might be emitted multiple times with the same `prompt` ID and the
```
reillyeon, line 1490:
```suggestion
event might be emitted multiple times with the same `promptID` and the
```

reillyeon (approved):

OrKoN:
@reillyeon @alexnj @nondebug is there anything blocking us from landing the spec change?

PR #630, opened 2024-08-16
Draft: Add Bluetooth device emulation support

alexnj:
Spec changes to support Bluetooth Device Emulation.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4951,15 +4951,6 @@ bluetooth.ScanRecord = {
 
 A bluetooth.ScanRecord represents data of the advertisement packet sent by a Bluetooth device.
 
-<div algorithm="serialize device">
-To <dfn>serialize a device</dfn> given a {{BluetoothDevice}} |device|:
-
-1. Let |name| be |device|.{{BluetoothDevice/name}}.
-1. Return a [=map=] matching the <code>bluetooth.RequestDeviceInfo</code> production, `"name"` set to |name|.
-
-</div>
-
-
 ### Errors ### {#bidi-errors}
 
 This specification extends the set of [=error codes=] from
@@ -5040,6 +5031,10 @@ A [=local end=] could dismiss a prompt by sending the following message:
 <pre highlight="cddl" class="cddl remote-cddl local-cddl">
 bluetooth.simulateCentral = (
    method: "bluetooth.simulateCentral",
+   params: bluetooth.SimulateCentralParameters,
+)
+
+bluetooth.SimulateCentralParameters = (
    state: "absent", "powered-off", or "powered-on"
 )
 </pre>
@@ -5068,6 +5063,10 @@ A [=local end=] could simulate a central by sending the following message:
 <pre highlight="cddl" class="cddl remote-cddl local-cddl">
 bluetooth.simulatePreconnectedPeripheral = (
    method: "bluetooth.simulatePreconnectedPeripheral",
+   params: bluetooth.SimulatePreconnectedPeripheralParameters,
+)
+
+bluetooth.SimulatePreconnectedPeripheralParameters = (
    address: string,
    name: string,
    manufacturerData: manufacturerData,
@@ -5078,18 +5077,26 @@ bluetooth.simulatePreconnectedPeripheral = (
 <div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
 The [=remote end steps=] with |command parameters| are:
 
-1. Let |state| be |params|[`"state"`].
+1. Let |address| be |params|[`"address"`].
+1. Let |name| be |params|[`"name"`].
+1. Let |manufacturerData| be |params|[`"manufacturerData"`].
+1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
 1. Return [=success=] with data `null`.
 
 </div>
 
 <div class="example">
-A [=local end=] could simulate a central by sending the following message:
+A [=local end=] could simulate a preconnected Peripheral by sending the following message:
 
 <pre highlight="json">
 {
   "method": "bluetooth.simulatePreconnectedPeripheral",
-  "state": "powered-on"
+  "address": "09:09:09:09:09:09",
+  "name": "Some Device",
+  "manufacturerData": [{ "companyIdentifier": 17 }],
+  "knownServiceUuids": [
+     "12345678-1234-5678-9abc-def123456789",
+  ]
 }
 </pre>
 </div>
@@ -5099,8 +5106,19 @@ A [=local end=] could simulate a central by sending the following message:
 <pre highlight="cddl" class="cddl remote-cddl local-cddl">
 bluetooth.simulateAdvertisement = (
    method: "bluetooth.simulateAdvertisement",
-   entry: ScanEntry[]
+   params: bluetooth.SimulateAdvertisementParameters,
 )
+
+bluetooth.SimulateAdvertisementParameters = (
+   entry: bluetooth.SimulateAdvertisementScanEntryParameters[]
+)
+
+bluetooth.SimulateAdvertisementScanEntryParameters = (
+   deviceAddress: string,
+   rssi: integer,
+   scanRecord: bluetooth.ScanRecord
+)
+
 </pre>
 
 <div algorithm="remote end steps for bluetooth.simulateAdvertisement">
@@ -5112,12 +5130,22 @@ The [=remote end steps=] with |command parameters| are:
 </div>
 
 <div class="example">
-A [=local end=] could simulate a central by sending the following message:
+A [=local end=] could simulate a device advertisement by sending the following message:
 
 <pre highlight="json">
 {
   "method": "bluetooth.simulateAdvertisement",
-  "entry": {}
+  "entry": {
+    "deviceAddress": "08:08:08:08:08:08",
+    "rssi": -10,
+    "scanRecord": {
+      "name": "Heart Rate",
+      "uuids": ["0000180d-0000-1000-8000-00805f9b34fb"],
+      "manufacturerData": [{ "companyIdentifier": 17 }],
+      "appearance": 1,
+      "txPower": 1
+    }
+  }
 }
 </pre>
 </div>
--- a/index.bs
+++ b/index.bs
@@ -4938,6 +4938,28 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
 
 </div>
 
+#### The bluetooth.ScanRecord Type #### {#bluetooth-scanrecord-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.ScanRecord = {
+  name: text / null,
+  uuids: [] / null,
+  appearance: integer / null,
+  manufacturerData: BluetoothManufacturerDataMap / null,
+}
+</pre>
+
+A bluetooth.ScanRecord represents data of the advertisement packet sent by a Bluetooth device.
+
+<div algorithm="serialize device">
+To <dfn>serialize a device</dfn> given a {{BluetoothDevice}} |device|:
+
+1. Let |name| be |device|.{{BluetoothDevice/name}}.
+1. Return a [=map=] matching the <code>bluetooth.RequestDeviceInfo</code> production, `"name"` set to |name|.
+
+</div>
+
+
 ### Errors ### {#bidi-errors}
 
 This specification extends the set of [=error codes=] from
@@ -5009,6 +5031,93 @@ A [=local end=] could dismiss a prompt by sending the following message:
     "accept": true,
     "device": "dvc-9b3b872"
   }
+}
+</pre>
+</div>
+
+#### The bluetooth.simulateCentral Command #### {#bluetooth-simulatecentral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulateCentral = (
+   method: "bluetooth.simulateCentral",
+   state: "absent", "powered-off", or "powered-on"
+)
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateCentral">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |state| be |params|[`"state"`].
+1. Return [=success=] with data `null`.
+
+</div>
+
+<div class="example">
+A [=local end=] could simulate a central by sending the following message:
+
+<pre highlight="json">
+{
+  "method": "bluetooth.simulateCentral",
+  "state": "powered-on"
+}
+</pre>
+</div>
+
+#### The bluetooth.simulatePreconnectedPeripheral Command #### {#bluetooth-simulateconnectedperipheral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulatePreconnectedPeripheral = (
+   method: "bluetooth.simulatePreconnectedPeripheral",
+   address: string,
+   name: string,
+   manufacturerData: manufacturerData,
+   knownServiceUuids: string[]
+)
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |state| be |params|[`"state"`].
+1. Return [=success=] with data `null`.
+
+</div>
+
+<div class="example">
+A [=local end=] could simulate a central by sending the following message:
+
+<pre highlight="json">
+{
+  "method": "bluetooth.simulatePreconnectedPeripheral",
+  "state": "powered-on"
+}
+</pre>
+</div>
+
+#### The bluetooth.simulateAdvertisement Command #### {#bluetooth-simulateadvertisement-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulateAdvertisement = (
+   method: "bluetooth.simulateAdvertisement",
+   entry: ScanEntry[]
+)
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with |command parameters| are:
+
+1. Let |entry| be |params|[`"entry"`].
+1. Return [=success=] with data `null`.
+
+</div>
+
+<div class="example">
+A [=local end=] could simulate a central by sending the following message:
+
+<pre highlight="json">
+{
+  "method": "bluetooth.simulateAdvertisement",
+  "entry": {}
 }
 </pre>
 </div>


alexnj:
@reillyeon @OrKoN First draft of BT device emulation support, would appreciate an early review and feedback. Some portions (like algorithm steps) are still missing, I'm working towards completing it.

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
 
+#### The bluetooth.simulateCentral Command #### {#bluetooth-simulatecentral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulateCentral = (
+   method: "bluetooth.simulateCentral",
+   params: bluetooth.SimulateCentralParameters,
+)
+
+bluetooth.SimulateCentralParameters = (
```
nondebug, line 5037: Use curly braces {} for map types like these parameters types.

```
bluetooth.SimulateCentralParameters = {
...
}
```

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+#### The bluetooth.simulatePreconnectedPeripheral Command #### {#bluetooth-simulateconnectedperipheral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulatePreconnectedPeripheral = (
+   method: "bluetooth.simulatePreconnectedPeripheral",
+   params: bluetooth.SimulatePreconnectedPeripheralParameters,
+)
+
+bluetooth.SimulatePreconnectedPeripheralParameters = (
```
nondebug, line 5069: Use curly braces for `SimulatePreconnectedPeripheralParameters`

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+#### The bluetooth.simulateAdvertisement Command #### {#bluetooth-simulateadvertisement-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulateAdvertisement = (
+   method: "bluetooth.simulateAdvertisement",
+   params: bluetooth.SimulateAdvertisementParameters,
+)
+
+bluetooth.SimulateAdvertisementParameters = (
```
nondebug, line 5112: Use curly braces for `SimulateAdvertisementParameters`

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   entry: bluetooth.SimulateAdvertisementScanEntryParameters[]
+)
+
+bluetooth.SimulateAdvertisementScanEntryParameters = (
```
nondebug, line 5116: Use curly braces for `SimulateAdvertisementScanEntryParameters`

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+<div class="example">
+A [=local end=] could simulate a preconnected Peripheral by sending the following message:
+
+<pre highlight="json">
+{
+  "method": "bluetooth.simulatePreconnectedPeripheral",
+  "address": "09:09:09:09:09:09",
+  "name": "Some Device",
+  "manufacturerData": [{ "companyIdentifier": 17 }],
```
nondebug, line 5096: I think square brackets [] are not needed here since we only want to pass one `BluetoothManufacturerDataMap`

```
"manufacturerData": { "companyIdentifier": 17 },
```

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+<pre highlight="json">
+{
+  "method": "bluetooth.simulateAdvertisement",
+  "entry": {
+    "deviceAddress": "08:08:08:08:08:08",
+    "rssi": -10,
+    "scanRecord": {
+      "name": "Heart Rate",
+      "uuids": ["0000180d-0000-1000-8000-00805f9b34fb"],
+      "manufacturerData": [{ "companyIdentifier": 17 }],
```
nondebug, line 5144:
```
"manufacturerData": { "companyIdentifier": 17 },
```

index.bs
```diff
@@ -4938,6 +4938,19 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
 </div>
 
+#### The bluetooth.ScanRecord Type #### {#bluetooth-scanrecord-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.ScanRecord = {
+  name: text / null,
+  uuids: [] / null,
+  appearance: integer / null,
+  manufacturerData: BluetoothManufacturerDataMap / null,
```
nondebug, line 4948: Consider making these optional instead of nullable

index.bs
```diff
@@ -4938,6 +4938,19 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
 
 </div>
 
+#### The bluetooth.ScanRecord Type #### {#bluetooth-scanrecord-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.ScanRecord = {
+  name: text / null,
+  uuids: [] / null,
```
nondebug, line 4946: `uuids` needs a type, perhaps `BluetoothServiceUuid[]`

nondebug (changes requested):

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
 </pre>
 </div>
 
+#### The bluetooth.simulateCentral Command #### {#bluetooth-simulatecentral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulateCentral = (
```
OrKoN, line 5032:
```suggestion
bluetooth.SimulateCentral = (
```

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+  "method": "bluetooth.simulateCentral",
+  "state": "powered-on"
+}
+</pre>
+</div>
+
+#### The bluetooth.simulatePreconnectedPeripheral Command #### {#bluetooth-simulateconnectedperipheral-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.simulatePreconnectedPeripheral = (
```
OrKoN, line 5064:
```suggestion
bluetooth.SimulatePreconnectedPeripheral = (
```

index.bs
```diff
@@ -5013,6 +5026,130 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+bluetooth.SimulateCentralParameters = {
+   state: "absent" / "powered-off" / "powered-on"
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateCentral">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |state| be |params|[`"state"`].
+1. Return [=success=] with data `null`.
```
OrKoN, line 5046: how it would be actually specified is not decided yet? what exactly the implementations should do to support the emulation commands?

index.bs
```diff
@@ -4850,6 +4850,8 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
+The web-bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver-BiDi extension commands that allow defining and controlling virtual peripherals and advertisements that behave like physical device peripherals and their advertisements. These virtual peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
```
reillyeon, line 4853:
```suggestion
The Web Bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver BiDi extension commands that allow defining and controlling virtual peripherals and advertisements that behave like physical device peripherals and their advertisements. These virtual peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
```

index.bs
```diff
@@ -4938,6 +4940,19 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
 
 </div>
 
+#### The bluetooth.ScanRecord Type #### {#bluetooth-scanrecord-type}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.ScanRecord = {
+  optional name: text,
+  optional uuids: BluetoothServiceUuid[],
+  optional appearance: integer,
```
reillyeon, line 4949: These fields should be documented, particularly ones like `appearance` where we can use `[=Appearance=]` to reference the appropriate part of the Bluetooth specifications.

index.bs
```diff
@@ -5013,6 +5028,114 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |address| be |params|[`"address"`].
+1. Let |name| be |params|[`"name"`].
+1. Let |manufacturerData| be |params|[`"manufacturerData"`].
+1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
+1. Let |virtualBluetoothDevice| be a new software defined {{BluetoothDevice}}.
```
reillyeon, line 5054: This should be a `[=Bluetooth device=]` rather than ` {{BluetoothDevice}}` to reference the abstract concept of a Bluetooth device, for which we set the properties below, but using their names from the [definition of a Bluetooth device](https://webbluetoothcg.github.io/web-bluetooth/#bluetooth-device).

index.bs
```diff
@@ -5013,6 +5028,114 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Set |virtualBluetoothDevice|'s name be |name|.
+1. Set |virtualBluetoothDevice|'s address be |address|.
+1. Set |virtualBluetoothDevice|'s manufacturer data be |manufacturerData|.
+1. Set |virtualBluetoothDevice|'s known service UUIDs be |knownServiceUuids|.
+1. Add |virtualBluetoothDevice| to <code>|storage|.{{BluetoothPermissionStorage/allowedDevices}}</code>.
```
reillyeon, line 5059: Doing this means the page doesn't have to call `navigator.bluetooth.requestDevice()`. Instead we need to add the `[=Bluetooth device=]` to a list which is checked by the `[=scan for devices=]` algorithm.

"Pre-connected" here means that the central already knows about the device without having to scan for it (e.g. the user has previously paired the device).

index.bs
```diff
@@ -5013,6 +5028,114 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+#### The bluetooth.simulateAdvertisement Command #### {#bluetooth-simulateadvertisement-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.SimulateAdvertisement = (
+   method: "bluetooth.simulateAdvertisement",
+   params: bluetooth.SimulateAdvertisementParameters,
+)
+
+bluetooth.SimulateAdvertisementParameters = {
+   entry: bluetooth.SimulateAdvertisementScanEntryParameters[]
```
reillyeon, line 5088: The steps below imply only a single scan event, but this is a list. I think only supporting one advertisement per call is fine, so just rename this to `scanEntry` and remove the "[]".

index.bs
```diff
@@ -5013,6 +5028,114 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let <var>device</var> be the virtual <a>Bluetooth device</a> that sent the
+    advertising event.
+1. For each {{BluetoothDevice}} <var>deviceObj</var> in the UA such that
+    <var>device</var> is the <a>same device</a> as
+    <code><var>deviceObj</var>.{{[[representedDevice]]}}</code>, <a>queue a
+    task</a> on <var>deviceObj</var>'s <a>relevant settings object</a>'s
+    <a>responsible event loop</a> to do the following sub-steps:
+    1. If <code><var>deviceObj</var>.{{watchingAdvertisements}}</code> is
+        `false`, abort these sub-steps.
+    1. <a>Fire an `advertisementreceived` event</a> for the advertising event at
```
reillyeon, line 5113: In addition to firing an `advertisementreceived` event we also want to inject the advertised device into the results from a concurrent call to the `[=scan for devices=]` algorithm.

reillyeon (changes requested):

index.bs
```diff
@@ -5081,8 +5051,13 @@ The [=remote end steps=] with command parameters |params| are:
 1. Let |name| be |params|[`"name"`].
 1. Let |manufacturerData| be |params|[`"manufacturerData"`].
 1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
+1. Let |virtualBluetoothDevice| be a new software defined {{BluetoothDevice}}.
```
nondebug, line 5054: virtualBluetoothDevice should be a [=Bluetooth device=] not a {{BluetoothDevice}}. In the spec, a [=Bluetooth device=] is a representation of a physical Bluetooth device with the properties defined in [5.1. Global Bluetooth device properties](https://webbluetoothcg.github.io/web-bluetooth/#global-device-properties).

Currently, section 5.1 is specific to physical Bluetooth devices. We may need to modify it to allow for virtual devices.

index.bs
```diff
@@ -5081,8 +5051,13 @@ The [=remote end steps=] with command parameters |params| are:
...
+1. Set |virtualBluetoothDevice|'s name be |name|.
+1. Set |virtualBluetoothDevice|'s address be |address|.
+1. Set |virtualBluetoothDevice|'s manufacturer data be |manufacturerData|.
+1. Set |virtualBluetoothDevice|'s known service UUIDs be |knownServiceUuids|.
+1. Add |virtualBluetoothDevice| to <code>|storage|.{{BluetoothPermissionStorage/allowedDevices}}</code>.
```
nondebug, line 5059: We don't want to add the device to allowedDevices as that would be equivalent to granting the device permission. Instead, we can [=get the BluetoothDevice representing=] |virtualBluetoothDevice| which will insert the device into [[deviceInstanceMap]] if the site is already allowed to access the device.

Calling [=get the BluetoothDevice representing=] also needs the context (navigator.bluetooth) which I think we can access with:

```
1. Let |traversable| be |session|'s [=current browsing context=]'s [=top-level traversable=].
1. Let |context| be |traversable|.{{Navigator/bluetooth}} if |traversable| is a {{Navigator}}; otherwise `null`.
```

index.bs
alexnj: Updated the section with notes with description. LMK if that's not the correct way.

index.bs
alexnj: I've added a `virtualDevices` list var into the scan algorithm, but I'm not confident of the way I've referenced in this section to crosslink. PTAL.

index.bs
```diff
@@ -1658,6 +1658,8 @@ following steps:
 1. Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
+1. Let <var>virtualDevices</var> be a set of simulated <a>Bluetooth device</a>s,
+    initially empty.
```
reillyeon, line 1662: We need to define what stores the set of simulated Bluetooth devices. See the example from [§ 9.1. Virtual Sensors](https://w3c.github.io/sensors/#virtual-sensors) where we give each "top-level traversable" (a fancy name for a tab) a virtual sensor mapping.

index.bs
```diff
@@ -110,6 +110,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
         text: error; url: dfn-error
         text: local end; url: dfn-local-ends
+        text: remote end steps; url: dfn-remote-end-steps
```
reillyeon, line 113: This and "error" above shouldn't need the "url" parameter because these are exported definitions. File a bug on the WebDriver spec to have "local end" exported so we can reference it here without hard-coding the URL.

index.bs
```diff
@@ -1658,6 +1659,12 @@ following steps:
 1. Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
+1. Let |topLevelTraversable| be <a for="request">window</a>'s [=Window/navigable=]'s
+    [=navigable/top-level traversable=].
+1. Let |virtualAdapter| be |topLevelTraversable|'s
+    <dfn>simulated Bluetooth adapter</dfn>, initially empty.
```
reillyeon, line 1665: "Simulated Bluetooth adapter" and "simulated Bluetooth device" should be defined somewhere outside these steps, probably in the introduction to the "Automated testing" section.

index.bs
```diff
@@ -1988,6 +1995,12 @@ steps <a>in parallel</a>:
       Note: If the Web Bluetooth permission has been blocked by the user, the UA
       may <a>resolve</a> |promise| with  `false`.
     </div>
+
+1. If [=this=]'s [=Window/navigable=]'s [=navigable/top-level traversable=] has
+    a <a>simulated Bluetooth adapter</a> set, <a>queue a
+    task</a> to <a>resolve</a> |promise| with `true` regardless of the powered
```
reillyeon, line 2001: This should check whether the simulated Bluetooth adapter's state is "absent" and return false in that case.

index.bs
```diff
@@ -4938,6 +4953,36 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+}
+</pre>
+
+A `bluetooth.ScanRecord` represents data of the advertisement packet sent by a [=Bluetooth device=].
+
+<div class="note" heading="bluetooth.ScanRecord attributes"
+  dfn-for="bluetooth.ScanRecord" dfn-type="attribute"
+  link-for-hint="bluetooth.ScanRecord">
+
+<dfn>name</dfn> is {{BluetoothDevice}}'s local name, or a prefix of it.
```
reillyeon, line 4973: Use `[=Bluetooth device=]` here because we're still talking about the abstract concept of a Bluetooth device that we're simulating an advertisement from and haven't created an event fired against an actual concrete `{{BluetoothDevice}}` instance (yet).

index.bs
```diff
@@ -4938,6 +4953,36 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+  optional name: text,
+  optional uuids: BluetoothServiceUuid[],
+  optional appearance: integer,
+  optional manufacturerData: BluetoothManufacturerDataMap,
+}
+</pre>
+
+A `bluetooth.ScanRecord` represents data of the advertisement packet sent by a [=Bluetooth device=].
+
+<div class="note" heading="bluetooth.ScanRecord attributes"
```
reillyeon, line 4969: Can you use a `<dl dfn-type=dict-member>` for this to get a bit better formatting?

See example: https://github.com/webmachinelearning/webnn/blob/main/index.bs#L1062

index.bs
```diff
@@ -5013,6 +5058,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   state: "absent" / "powered-off" / "powered-on"
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdapter">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |state| be |params|[`"state"`].
+1. Let |virtualBluetoothAdapter| be a new [=simulated Bluetooth adapter=].
+1. Set |virtualBluetoothAdapter|'s state be |state|.
```
reillyeon, line 5079: When we define what a "simulated Bluetooth adapter" is that should include a definition of what the possible state values are, similar to how the definition of a "[virtual sensor](https://w3c.github.io/sensors/#virtual-sensors)" provides a list of all the associated data. This would also include things like the list of simulated Bluetooth devices.

index.bs
```diff
@@ -5013,6 +5058,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+1. Let |address| be |params|[`"address"`].
+1. Let |name| be |params|[`"name"`].
+1. Let |manufacturerData| be |params|[`"manufacturerData"`].
+1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
+1. Let |virtualBluetoothDevice| be a new [=simulated Bluetooth device=].
+1. Set |virtualBluetoothDevice|'s name be |name|.
+1. Set |virtualBluetoothDevice|'s address be |address|.
+1. Set |virtualBluetoothDevice|'s manufacturer data be |manufacturerData|.
+1. Set |virtualBluetoothDevice|'s known service UUIDs be |knownServiceUuids|.
```
reillyeon, line 5126: Similar to the above, these should be defined properties for a "simulated Bluetooth device" with links to their definitions similar to the virtual sensor example.

reillyeon (changes requested):

index.bs
nondebug: See https://w3c.github.io/sensors/#virtual-sensors in the Sensors API spec for an example.

index.bs
```diff
@@ -5013,6 +5058,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |scanEntry| be |params|[`"scanEntry"`].
+1. Let <var>device</var> be the virtual <a>Bluetooth device</a> that sent the
+    advertising event.
+1. For each {{BluetoothDevice}} <var>deviceObj</var> in the UA such that
```
nondebug, line 5174: "in the UA" presumably means all the BluetoothDevice values in [[deviceInstanceMap]]. I see we're already using this pattern elsewhere in the spec but I think it will make it clearer if we explicitly loop over [[deviceInstanceMap]]. Here's my attempt to get the map via the Bluetooth object:

```
The [=remote end steps=] with |session| and command parameters |params| are:

1. Let |global| be |session|'s [=current browsing context=]'s [=browsing context/active document=]'s [=relevant global object=].
1. If |global| is not a {{Window}}, abort these steps.
1. Let |navigator| be |global|'s associated {{Navigator}}.
1. Let |deviceObjects| be the result of [=map/get the values=] for |navigator|.{{Navigator/bluetooth}}.{{Bluetooth/[[deviceInstanceMap]]}}.
1. [=list/For each=] |deviceObject| of |deviceObjects| ...
```

index.bs
```diff
@@ -5013,6 +5058,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |name| be |params|[`"name"`].
+1. Let |manufacturerData| be |params|[`"manufacturerData"`].
+1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
+1. Let |virtualBluetoothDevice| be a new [=simulated Bluetooth device=].
+1. Set |virtualBluetoothDevice|'s name be |name|.
+1. Set |virtualBluetoothDevice|'s address be |address|.
+1. Set |virtualBluetoothDevice|'s manufacturer data be |manufacturerData|.
+1. Set |virtualBluetoothDevice|'s known service UUIDs be |knownServiceUuids|.
+1. Add |virtualBluetoothDevice| to <code>virtualDevices</code> under [=scan for devices=].
+1. Return [=success=] with data `null`.
```
nondebug, line 5128: There are probably some things we want to validate before the device is added.

What should happen if the device address matches an existing device, virtual or non-virtual?
Can name, address, manufacturerData, or knownServiceUuids be empty?

index.bs
```diff
@@ -4850,6 +4863,8 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
+The Web Bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver-BiDi extension commands that allow defining and controlling virtual peripherals and advertisements that behave like physical device peripherals and their advertisements. These virtual peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
```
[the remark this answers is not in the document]
alexnj: Picked simulated

alexnj (commented):
Tackled some of the feedback.

index.bs
```diff
@@ -5013,6 +5058,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   scanRecord: bluetooth.ScanRecord
+}
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |scanEntry| be |params|[`"scanEntry"`].
+1. Let <var>device</var> be the virtual <a>Bluetooth device</a> that sent the
```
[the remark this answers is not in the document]
alexnj: I think this should be taken care of now — the `device` is matched against each in `deviceInstanceMap` for equality, and advertisement is sent only against the matched. Would that work?

index.bs
```diff
@@ -1658,6 +1664,11 @@ following steps:
 1. Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
+1. Let |topLevelTraversable| be <a for="request">window</a>'s [=Window/navigable=]'s
+    [=navigable/top-level traversable=].
```
reillyeon, line 1668: `<a for="request">window</a>` is an unrelated reference to the Fetch API. We probably want to pass a _global_ parameter to the "scan for devices" algorithm and then say,

```suggestion
1. Let |topLevelTraversable| be |global|'s [=Window/navigable=]'s [=navigable/top-level
        traversable=].
```

Or we could pass a top-level traversable directly. That might align better with the text below which says to look for "scan for devices" requests for a given top-level traversable.

index.bs
```diff
@@ -1658,6 +1664,11 @@ following steps:
...
+1. Let |simulatedBluetoothAdapter| be |topLevelTraversable|'s
+    <a>simulated Bluetooth adapter</a>, initially empty.
+1. Let |simulatedBluetoothDevices| be a set of |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device</a>s.
```
reillyeon, line 1671: We don't need to say "initially empty" because that's explained when we define the concept of a simulated adapter but we do have to handle the case where there isn't a simulated adapter. I would say,

```suggestion
1. Let |simulatedBluetoothDevices| be an empty [=set=].
1. If |topLevelTraversable| has a [=simulated Bluetooth adapter=], let |simulatedBluetoothDevices| be its set of [=simulated Bluetooth devices=].
```

Note that Bikeshed can [handle plurals automatically](https://speced.github.io/bikeshed/#conjugating).

index.bs
```diff
@@ -1988,6 +1999,16 @@ steps <a>in parallel</a>:
...
     </div>
+
+1. Let |simulatedBluetoothAdapter| be [=this=]'s [=Window/navigable=]'s
+    [=navigable/top-level traversable=] <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is not empty, and
+    |simulatedBluetoothAdapter|["state"] is equal to "absent", <a>queue a
+    task</a> to <a>resolve</a> |promise| with `false`.
+1. If |simulatedBluetoothAdapter| is not empty, regardless of the powered
+    state of the <a>simulated Bluetooth adapter</a>, <a>queue a
+    task</a> to <a>resolve</a> |promise| with `true`.
```
reillyeon, line 2010:
```suggestion
1. Let |simulatedBluetoothAdapter| be [=this=]'s [=Window/navigable=]'s
    [=navigable/top-level traversable=] <a>simulated Bluetooth adapter</a>.
1. If |simulatedBluetoothAdapter| is not empty,
    1. If |simulatedBluetoothAdapter|["state"] is "absent", [=queue a task=] to [=resolve=] |promise| with `false`.
    1. Otherwise, [=queue a task=] to [=resolve=] |promise| with `true`.
    1. Abort these steps.
```

index.bs
```diff
@@ -4938,6 +4965,34 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+</pre>
+
+A `bluetooth.ScanRecord` represents data of the advertisement packet sent by a [=Bluetooth device=].
+
+<dl>
+  <dt><code>name</code></dt>
+  <dd>is {{BluetoothDevice}}'s local name, or a prefix of it.</dd>
+
+  <dt><code>uuids</code></dt>
+  <dd>lists the Service UUIDs that this scan record says {{BluetoothDevice}}'s GATT server supports.</dd>
```
reillyeon, line 4986:
```suggestion
  <dd>lists the Service UUIDs that this scan record says the [=Bluetooth device=]'s GATT server supports.</dd>
```

index.bs
```diff
@@ -4938,6 +4965,34 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+
+  <dt><code>appearance</code></dt>
+  <dd>is an <a>Appearance</a>, one of the values defined by the {{gap.appearance}} characteristic.</dd>
+
+  <dt><code>manufacturerData</code></dt>
+  <dd>maps <code>unsigned short</code> Company Identifier Codes to {{DataView}}s.</dd>
```
reillyeon, line 4992: In the WebDriver API it's not going to be a DataView (that's a JavaScript type). It probably needs to be a simple list of unsigned octets.

index.bs
```diff
@@ -5013,6 +5068,161 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with |session| and command parameters |params| are:
+
+1. Let |address| be |params|[`"address"`].
+1. Let |name| be |params|[`"name"`].
+1. Let |manufacturerData| be |params|[`"manufacturerData"`].
+1. Let |knownServiceUuids| be |params|[`"knownServiceUuids"`].
```
reillyeon, line 5130: It's probably more readable to skip defining these variables and just directly assign properties of _simulatedBluetoothDevice_ to the values in _params_.

index.bs
```diff
@@ -5013,6 +5068,161 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+<div class="example">
+A [=local end=] could simulate a preconnected peripheral by sending the following message:
+
+<pre highlight="json">
+{
+  "method": "bluetooth.simulatePreconnectedPeripheral",
+  "address": "09:09:09:09:09:09",
+  "name": "Some Device",
+  "manufacturerData": { "companyIdentifier": 17 },
```
reillyeon, line 5151: This example is missing the actual data associated with the company ID. It would probably be something like,

```suggestion
  "manufacturerData": [{ "companyIdentifier": 17, "data": [0, 255, 1, 1, 127] }],
```

or

```suggestion
  "manufacturerData": { 17: [0, 255, 1, 1, 127] },
```

index.bs
```diff
@@ -5013,6 +5068,161 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   scanRecord: bluetooth.ScanRecord
+}
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with |session| and command parameters |params| are:
+
+1. Let |scanEntry| be |params|[`"scanEntry"`].
+1. Let <var>device</var> be the <a>simulated Bluetooth device</a> that sent the
```
reillyeon, line 5183: Look up the simulated Bluetooth device attached to _session_'s simulated Bluetooth adapter with the given _deviceAddress_.

index.bs
```diff
@@ -5013,6 +5068,161 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+<pre highlight="json">
+{
+  "method": "bluetooth.simulateAdvertisement",
+  "entry": {
+    "deviceAddress": "08:08:08:08:08:08",
+    "rssi": -10,
+    "scanRecord": {
+      "name": "Heart Rate",
+      "uuids": ["0000180d-0000-1000-8000-00805f9b34fb"],
+      "manufacturerData": { "companyIdentifier": 17 },
```
reillyeon, line 5217: Same issue here with manufacturer data.

reillyeon (changes requested):

index.bs
alexnj: To be on the same page, we need to change the `dd` here so it reads `to a list of unsigned octets.`?

index.bs
alexnj: We're currently not doing any validation (other than it being a map, keyed by device address). All the fields are defined optional.

index.bs
```diff
@@ -5013,6 +5066,155 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |topLevelTraversable| be |session|'s [=current browsing context=]'s [=navigable/top-level traversable=].
+1. Let |simulatedBluetoothAdapter| be |topLevelTraversable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, return [=error=].
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping| does not have a <a>simulated Bluetooth device</a> with its Bluetooth address equal to |deviceAddress|, create the <a>simulated Bluetooth device</a> with |deviceAddress| and insert it into |deviceMapping|.
+1. Let |deviceObject| be the [=bluetooth device=] in |deviceMapping| where |deviceObject|'s device address is equal to |deviceAddress|.
+1. <a>Queue a task</a> on <var>deviceObject</var>'s <a>relevant settings object</a>'s <a>responsible event loop</a> to do the following sub-steps:
+    1. If the [=navigable/top-level traversable=] is currently executing a
+        [=scan for devices=], insert <var>deviceObject</var> into
+        [=simulated Bluetooth device=]s.
```
reillyeon, line 5188: This is always going to be hand-wavy until we improve the scan for devices algorithm but I think I would say,

```suggestion
    1. If the [=navigable/top-level traversable=] is currently executing the
        [=scan for devices=] algorithm, insert <var>simulatedDevice</var> into
        the <em>simulatedBluetoothDevices</em> variable within that algorithm.
```

We need a new _simulatedDevice_ variable which is the simulated Bluetooth device created or looked up by the steps above so that we are inserting something of the right type into the _simulatedBluetoothDevices_ variable in the scan for devices algorithm steps. Defining this as a separate variable will improve the readability of the steps above anyways.

We should add a note acknowledging that inserting data into variables from another algorithm is extremely hand-wavey and reference the issue tracking improving that algorithm to support asynchronous device discovery.

reillyeon (changes requested):

index.bs
```diff
@@ -4850,6 +4867,14 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
+The Web Bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver-BiDi extension commands that allow defining and controlling simulated peripherals and advertisements that behave like physical device peripherals and their advertisements. These simulated peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
+
+Each [=navigable/top-level traversable=] may have a <dfn>simulated Bluetooth adapter</dfn>, that is a software defined Bluetooth adapter that has a set of discovered <a>simulated Bluetooth device</a>s and can assume roles like <a>Central</a>.
+
+Each <a>simulated Bluetooth adapter</a> has a <dfn>simulated Bluetooth device mapping</dfn>, which is a map of Bluetooth address to <a>simulated Bluetooth device</a>.
```
reillyeon, line 4874:
```suggestion
Each <a>simulated Bluetooth adapter</a> has a <dfn>simulated Bluetooth device mapping</dfn>, which is an <a>ordered map</a> of Bluetooth addresss to <a>simulated Bluetooth devices</a>.
```

index.bs
```diff
@@ -4850,6 +4867,14 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
+The Web Bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver-BiDi extension commands that allow defining and controlling simulated peripherals and advertisements that behave like physical device peripherals and their advertisements. These simulated peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
+
+Each [=navigable/top-level traversable=] may have a <dfn>simulated Bluetooth adapter</dfn>, that is a software defined Bluetooth adapter that has a set of discovered <a>simulated Bluetooth device</a>s and can assume roles like <a>Central</a>.
```
reillyeon, line 4872:
```suggestion
Each [=navigable/top-level traversable=] may have a <dfn>simulated Bluetooth adapter</dfn>, that is a software defined Bluetooth adapter that has a set of discovered <a>simulated Bluetooth devices</a> and can assume roles like <a>Central</a>.
```

reillyeon (changes requested):

index.bs
```diff
@@ -5181,15 +5184,17 @@ The [=remote end steps=] with |session| and command parameters |params| are:
 1. If |simulatedBluetoothAdapter| is empty, return [=error=].
 1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
 1. If |deviceMapping| does not have a <a>simulated Bluetooth device</a> with its Bluetooth address equal to |deviceAddress|, create the <a>simulated Bluetooth device</a> with |deviceAddress| and insert it into |deviceMapping|.
-1. Let |deviceObject| be the [=bluetooth device=] in |deviceMapping| where |deviceObject|'s device address is equal to |deviceAddress|.
-1. <a>Queue a task</a> on <var>deviceObject</var>'s <a>relevant settings object</a>'s <a>responsible event loop</a> to do the following sub-steps:
+1. Let |simulatedDevice| be the [=bluetooth device=] in |deviceMapping| where |simulatedDevice|'s device address is equal to |deviceAddress|.
+1. <a>Queue a task</a> on <var>simulatedDevice</var>'s <a>relevant settings object</a>'s <a>responsible event loop</a> to do the following sub-steps:
```
reillyeon, line 5188: _simulatedDevice_ does not have a relevant settings object because it's a "Bluetooth device", which is an abstract concept, and not a `{{BluetoothDevice}}`. There's a missing step here which looks up the correct `{{BluetoothDevice}}` in `navigable.bluetooth.[[deviceInstanceMap]]`.

index.bs
```diff
@@ -5181,15 +5184,17 @@ The [=remote end steps=] with |session| and command parameters |params| are:
...
     1. If the [=navigable/top-level traversable=] is currently executing the
```
reillyeon, line 5189: Pull this step out of the queued task. Since it relates to what is effectively browser UI it doesn't need to happen on any particular context's event loop and there isn't a good way to look one up to begin with.

index.bs
alexnj: Added deviceAddress validation before adding.

OrKoN (approved):
LGTM (let's verify that CDDL compiles as discussed offline)

index.bs
```diff
@@ -5013,6 +5073,161 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   name: string,
+   manufacturerData: BluetoothManufacturerDataMap,
+   knownServiceUuids: string[]
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with |session| and command parameters |params| are:
+
+1. Let |topLevelTraversable| be |session|'s [=current browsing context=]'s [=navigable/top-level traversable=].
```
OrKoN, line 5132: WebDriver BiDi session does not have a concept of the current browsing context since multiple contexts can be operated with at the same time.

We need to accept a `context: string` argument in this command if it needs to be frame-specific.  https://w3c.github.io/webdriver-bidi/#get-a-navigable can be used to resolve the id to navigable.

Also, it needs to be considered whether it should be supported for inner frames individually or only for the top-level frame and whether it applies to descendats.

alexnj: Thanks for the input! Revised steps to reflect resolution of navigable via context param. PTAL!

index.bs
```diff
@@ -2495,8 +2516,7 @@ broadcasting, measured in dBm. This is used to compute the path loss as
 in dBm. This is used to compute the path loss as <code>this.txPower -
 this.rssi</code>.
 
-<dfn>manufacturerData</dfn> maps <code>unsigned short</code> Company Identifier
-Codes to {{DataView}}s.
+<dfn>manufacturerData</dfn> maps <code>unsigned short</code> Company IdentifierCodes to list of unsigned octets.
```
reillyeon, line 2519: I think this is still true for the API exposed to script, but we need a separate definition of the manufacturer data map for the WebDriver commands.

index.bs
```diff
@@ -4827,6 +4847,10 @@ if the following steps return `blocked`:
 
 # Extensions to the Navigator Interface # {#navigator-extensions}
 
+Each {{Window}} has an <dfn>associated `Bluetooth`</dfn>, which is a {{Bluetooth}} object. Upon creation of the {{Window}} object, its <a>associated `Bluetooth`</a> must be set to a new {{Bluetooth}} object created in the {{Window}} object's [=relevant realm=].
```
reillyeon, line 4850: Move this new text to after the WebIDL below.

index.bs
```diff
@@ -4827,6 +4847,10 @@ if the following steps return `blocked`:
...
+
+{{Navigator}}'s {{bluetooth}} getter steps are to return [=this=]'s <a>associated `Bluetooth`</a>.
```
reillyeon, line 4852: This should be `<dfn for="Navigator">bluetooth</dfn>` so that the WebIDL block links here.

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+
+bluetooth.SimulateAdapterParameters = {
+   state: "absent" / "powered-off" / "powered-on"
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdapter">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let contextId be params["context"].
```
reillyeon, line 5092: Add `context` to `bluetooth.SimulateAdapterParameters`.

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   name: string,
+   manufacturerData: BluetoothManufacturerDataMap,
+   knownServiceUuids: string[]
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be params["context"].
```
reillyeon, line 5133: Update the parameter struct here too.

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+   rssi: integer,
+   scanRecord: bluetooth.ScanRecord
+}
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be params["context"].
```
reillyeon, line 5190: Update the parameter struct here too.

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. If the [=navigable/top-level traversable=] is currently executing the
+    [=scan for devices=] algorithm, insert <var>simulatedDevice</var> into
+    the <em>simulatedBluetoothDevices</em> variable within that algorithm.
+
+    Issue: Inserting data into variables from another algorithm could be improved.
+1. Let |navigables| be the <a>inclusive descendant navigables</a> of |topLevelNavigable|'s <a>active document</a>.
+1. For each |navigable| of |navigables|:
+    1. Let |document| be |navigable|'s <a>active document</a>.
+    1. <a>Queue a task</a> on |document|'s <a>relevant settings object</a>'s <a>responsible event loop</a> to do the following sub-steps:
+        1. Let |simulatedDeviceInstance| be the {{BluetoothDevice}} the |navigable|'s `navigator.bluetooth.[[deviceInstanceMap]]` corresponding to |simulatedDevice|.
```
reillyeon, line 5208: You still need to do the explicit walk through the active document, associated Navigator and associated Bluetooth to get the object you can actually look up [[deviceInstanceMap]] on.

reillyeon (changes requested):

index.bs
```diff
@@ -4938,6 +4970,34 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+  optional appearance: integer,
+  optional manufacturerData: BluetoothManufacturerDataMap,
+}
+</pre>
+
+A `bluetooth.ScanRecord` represents data of the advertisement packet sent by a [=Bluetooth device=].
+
+<dl>
+  <dt><code>name</code></dt>
+  <dd>is {{BluetoothDevice}}'s local name, or a prefix of it.</dd>
```
nondebug, line 4988:
```
<dd>is the [=Bluetooth device=]'s local name, or a prefix of it.</dd>
```

index.bs
nondebug: maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octets}}.

index.bs
nondebug, line 5092: nit: |contextId|

```
1. Let |contextId| be params["context"].
```

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+}
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateAdapter">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let contextId be params["context"].
+1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
+1. If |navigable| is not a [=top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |state| be |params|[`"state"`].
```
nondebug, line 5095: The local |state| doesn't seem to be needed, we can set |simulatedBluetoothAdapter|'s state to |params|[`"state"`]

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |simulatedBluetoothAdapter| be a new [=simulated Bluetooth adapter=].
+1. Set |simulatedBluetoothAdapter|'s state to |state|.
```
nondebug, line 5097: Please update the [=simulated Bluetooth adapter=] definition to include that a simulated adapter has a state.

index.bs
```diff
@@ -4850,6 +4874,14 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
 
 For the purposes of user-agent automation and application testing, this document defines extensions to the [[WebDriver-BiDi]] specification.
 
+The Web Bluetooth API and its extension specifications pose a challenge to test authors, as fully exercising those interfaces requires physical hardware devices that respond in predictable ways. To address this challenge this document defines a number of WebDriver-BiDi extension commands that allow defining and controlling simulated peripherals and advertisements that behave like physical device peripherals and their advertisements. These simulated peripherals and advertisements represent devices with particular properties and whose readings can be entirely defined by users.
+
+Each [=navigable/top-level traversable=] may have a <dfn>simulated Bluetooth adapter</dfn>, that is a software defined Bluetooth adapter that has a set of discovered <a>simulated Bluetooth devices</a> and can assume roles like <a>Central</a>.
+
+Each <a>simulated Bluetooth adapter</a> has a <dfn>simulated Bluetooth device mapping</dfn>, which is an <a>ordered map</a> of Bluetooth address to <a>simulated Bluetooth devices</a>.
```
nondebug, line 4881: This should specify that the key type is a string so we can index into it with the string we get from WebDriver BiDi.

```
which is an <a>ordered map</a> of Bluetooth address {{strings}} to <a>simulated Bluetooth devices</a>.
```

index.bs
nondebug, line 5190: Add a context parameter to bluetooth.SimulateAdvertisementParameters

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulatePreconnectedPeripheral">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be params["context"].
+1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
+1. If |navigable| is not a [=top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |simulatedBluetoothAdapter| be |navigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, return [=error=].
```
nondebug, line 5137: [=error=] should specify which error code to return.

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |deviceAddress| be |params|[`"address"`].
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping|[|deviceAddress|] exists, return [=error=].
```
nondebug, line 5140: Specify error code. Also, we can link [=exists=]

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+<div algorithm="remote end steps for bluetooth.simulateAdvertisement">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be params["context"].
+1. Let |topLevelNavigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
+1. If |topLevelNavigable| is not a [=top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |scanEntry| be |params|[`"scanEntry"`].
+1. Let |deviceAddress| be |scanEntry|[`"deviceAddress"`].
+1. Let |simulatedBluetoothAdapter| be |topLevelNavigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, return [=error=].
```
nondebug, line 5196: Specify error code

index.bs
```diff
@@ -5013,6 +5073,166 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping|[|deviceAddress|] exists, let |simulatedDevice| be |deviceMapping|[|deviceAddress|]. Otherwise, let |simulatedDevice| be a new <a>simulated Bluetooth device</a> with |deviceAddress| and set |deviceMapping|[|deviceAddress|] to |simulatedDevice|.
```
nondebug, line 5198: nit: [=exists=]

index.bs
nondebug, line 5208: nit: Add "in"

```
1. Let |simulatedDeviceInstance| be the {{BluetoothDevice}} in the |navigable|'s `navigator.bluetooth.[[deviceInstanceMap]]` corresponding to |simulatedDevice|.
```

alexnj:
```
1. Let |simulatedDeviceInstance| be the result of 
<a>get the <code>BluetoothDevice</code> representing</a> 
|simulatedDevice| inside |navigable|'s <a>active window</a>'s 
[=associated Bluetooth=].
```

index.bs
```diff
@@ -1415,8 +1421,8 @@ steps:
   devices, but for now it only ever returns a single one.
 </div>
 
-1. Let |document| be [=this=]'s [=relevant global object=]'s
-    [=associated Document=].
+1. Let |global| be [=this=]'s [=relevant global object=].
```
reillyeon, line 1424: Similar to the permission query algorithm this should probably say "Let _global_ be the <a>relevant global object for _storage_."

index.bs
```diff
@@ -1476,7 +1482,7 @@ steps:
     for example because there is no Bluetooth adapter with which to scan, or
     because the filters can't be matched by any possible advertising packet, the
     UA MAY return `[]` and abort these steps.
-1. <a>Scan for devices</a> with <var>requiredServiceUUIDs</var> as the <i>set of
+1. <a>Scan for devices</a> with |global| as the <i>relevant global object</i> and <var>requiredServiceUUIDs</var> as the <i>set of
     <a>Service</a> UUIDs</i>, and let <var>scanResult</var> be the result.
```
reillyeon, line 1486: We can simplify this to:

```suggestion
1. Let |scanResult| be the result of invoking [=scan for devices=] with |global| and
    |requiredServiceUUIDs|.
```

index.bs
```diff
@@ -1647,7 +1653,7 @@ returned from the following steps:
 </div>
 
 <div algorithm="scanning for Bluetooth devices">
-To <dfn>scan for devices</dfn> with an optional <var>set of <a>Service</a>
+To <dfn>scan for devices</dfn> with parameters <var>relevant global object</var> and an optional <var>set of <a>Service</a>
```
reillyeon, line 1656:
```suggestion
To <dfn>scan for devices</dfn> with parameters <var>global</var> and an optional <var>set of <a>Service</a>
```

index.bs
```diff
@@ -1658,6 +1664,14 @@ following steps:
 1. Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
+1. Let |topLevelTraversable| be the |relevant global object|'s [=Window/navigable=]'s
```
reillyeon, line 1667:
```suggestion
1. Let |topLevelTraversable| be the |global|'s [=Window/navigable=]'s
```

index.bs
```diff
@@ -2495,8 +2517,7 @@ broadcasting, measured in dBm. This is used to compute the path loss as
 in dBm. This is used to compute the path loss as <code>this.txPower -
 this.rssi</code>.
 
-<dfn>manufacturerData</dfn> maps <code>unsigned short</code> Company Identifier
-Codes to {{DataView}}s.
+<dfn>manufacturerData</dfn> maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octets}}.
```
reillyeon, line 2520: In the JavaScript API the map value is a `DataView`.

index.bs
```diff
@@ -5013,6 +5087,177 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. Let |contextId| be params["context"].
+1. Let |topLevelNavigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
+1. If |topLevelNavigable| is not a [=top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |scanEntry| be |params|[`"scanEntry"`].
+1. Let |deviceAddress| be |scanEntry|[`"deviceAddress"`].
+1. Let |simulatedBluetoothAdapter| be |topLevelNavigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, return [=error=] with [=error code=] [=invalid argument=].
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping|[|deviceAddress|] [=exists=], let |simulatedDevice| be |deviceMapping|[|deviceAddress|]. Otherwise, let |simulatedDevice| be a new <a>simulated Bluetooth device</a> with |deviceAddress| and set |deviceMapping|[|deviceAddress|] to |simulatedDevice|.
+1. If the [=navigable/top-level traversable=] is currently executing the
```
reillyeon, line 5221:
```suggestion
1. If |topLevelTraversable| is currently executing the
```

index.bs
```diff
@@ -5013,6 +5087,177 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+    [=scan for devices=] algorithm, insert <var>simulatedDevice</var> into
+    the <em>simulatedBluetoothDevices</em> variable within that algorithm.
+
+    Issue: Inserting data into variables from another algorithm could be improved.
```
reillyeon, line 5225:
```suggestion
    Issue: Inserting data into variables from another algorithm is not well defined. The <a>scan for devices</a> algorithm needs to define asynchronous device discovery in order to match implementations.
```

reillyeon (changes requested):

index.bs
alexnj: I think it should be |topLevelNavigable|, as |topLevelTraversable| isn't defined in this algorithm.

index.bs
```diff
@@ -1658,6 +1664,14 @@ following steps:
 1. Let <var>nearbyDevices</var> be a set of <a>Bluetooth device</a>s, initially
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
+1. Let |topLevelTraversable| be the |relevant global object|'s [=Window/navigable=]'s
+    [=navigable/top-level traversable=].
+1. Let |simulatedBluetoothDevices| be an empty [=set=].
```
alexnj, line 1669: This probably should be [=list=]?

index.bs
```diff
@@ -1664,11 +1664,11 @@ following steps:
...
     equal to the set of devices that are connected (have an <a>ATT Bearer</a>)
     to the UA.
-1. Let |topLevelTraversable| be the |relevant global object|'s [=Window/navigable=]'s
+1. Let |topLevelTraversable| be the |global|'s [=Window/navigable=]'s
     [=navigable/top-level traversable=].
-1. Let |simulatedBluetoothDevices| be an empty [=set=].
+1. Let |simulatedBluetoothDevices| be an empty <a spec="infra">list</a>.
 1. If |topLevelTraversable| has a [=simulated Bluetooth adapter=], let
-    |simulatedBluetoothDevices| be its set of [=simulated Bluetooth devices=].
+    |simulatedBluetoothDevices| be the result of [=get the values=] of its  [=simulated Bluetooth device mapping=].
```
reillyeon, line 1671:
```suggestion
    |simulatedBluetoothDevices| be the result of [=getting the values=] of its  [=simulated Bluetooth device mapping=].
```

index.bs
```diff
@@ -4858,7 +4859,7 @@ if the following steps return `blocked`:
 
 Each {{Navigator}} has an <dfn>associated `Bluetooth`</dfn>, which is a {{Bluetooth}} object. Upon creation of the {{Navigator}} object, its <a>associated `Bluetooth`</a> must be set to a new {{Bluetooth}} object created in the {{Navigator}} object's [=relevant realm=].
 
-{{Navigator}}'s <dfn for="Navigator">bluetooth</dfn> getter steps are to return [=this=]'s <a>associated `Bluetooth`</a>.
+{{Navigator}}'s {{Navigator/bluetooth}} getter steps are to return [=this=]'s <a>associated `Bluetooth`</a>.
```
reillyeon, line 4862: Why this change? I think using `<dfn for="Navigator">bluetooth</dfn>` is correct here as this is the definition for this attribute.

alexnj: It was giving a warning saying that it's an unused definition. Also, the IDL reference is the getter method, right?

reillyeon: I think you need to add the `attribute` attribute to the `<dfn>` tag for it to recognize that we're providing a more precise definition for something from the WebIDL block.

alexnj: TIL! Thanks, that resolved it.

alexnj:
@reillyeon @nondebug is this good to merge?

index.bs
```diff
@@ -5013,6 +5087,177 @@ A [=local end=] could dismiss a prompt by sending the following message:
...
+1. If |navigable| is not a [=top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |simulatedBluetoothAdapter| be |navigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, return [=error=] with [=error code=] [=invalid argument=].
+1. Let |deviceAddress| be |params|[`"address"`].
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping|[|deviceAddress|] [=exists=], return [=error=] with [=error code=] [=invalid argument=].
+1. Let |simulatedBluetoothDevice| be a new [=simulated Bluetooth device=].
+1. Set |simulatedBluetoothDevice|'s name to |params|[`"name"`].
+1. Set |simulatedBluetoothDevice|'s address to |params|[`"address"`].
+1. Set |simulatedBluetoothDevice|'s manufacturer data to |params|[`"manufacturerData"`].
```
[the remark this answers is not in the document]
alexnj: Is this an issue in the base spec that needs to be tackled at some point, or a deviation from current implementation of peripherals? It looks like the current implementation does associate manufacturerData with peripherals on creation, but the spec has a notion of manufacturer data only as part of advertisement. Curious how manufacturer data association with a simulated peripheral would work (it does match the implementation though)

index.bs
```diff
@@ -4938,6 +4983,35 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+  <dd>is the [=Bluetooth device=]'s local name, or a prefix of it.</dd>
+
+  <dt><code>uuids</code></dt>
+  <dd>lists the Service UUIDs that this scan record says the [=Bluetooth device=]'s GATT server supports.</dd>
+
+  <dt><code>appearance</code></dt>
+  <dd>is an <a>Appearance</a>, one of the values defined by the {{gap.appearance}} characteristic.</dd>
+
+  <dt><code>manufacturerData</code></dt>
+  <dd>maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octets}}.</dd>
```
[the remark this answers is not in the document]
alexnj: The CDDL is correct, the description seems mismatched. I can think of two ways of addressing this:
```
lists [=maplike=] objects that maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octets}}.
```

```
list of <code>bluetooth.BluetoothManufacturerData</code> that maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octets}}.
```

index.bs
```diff
@@ -4938,6 +4984,35 @@ To <dfn>serialize prompt devices</dfn> given [=device prompt=] |prompt|:
...
+  <dd>is the [=Bluetooth device=]'s local name, or a prefix of it.</dd>
+
+  <dt><code>uuids</code></dt>
+  <dd>lists the Service UUIDs that this scan record says the [=Bluetooth device=]'s GATT server supports.</dd>
+
+  <dt><code>appearance</code></dt>
+  <dd>is an <a>Appearance</a>, one of the values defined by the {{gap.appearance}} characteristic.</dd>
+
+  <dt><code>manufacturerData</code></dt>
+  <dd>maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octet}}s.</dd>
```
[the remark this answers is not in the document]
alexnj:
```
LINK ERROR: No 'idl' refs found for 'octets'.
{{octets}}
```

IIRC thats how I ended up writing as {{octet}}s. May be there is an alternative?

index.bs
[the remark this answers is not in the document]
alexnj: I've updated it to:
(1) add a property definition in the simulated device definition. This still felt incomplete to me as a proper definition, so I linked it to the reference where manufacturer data and service UUIDs are referenced elsewhere in the spec.
(2) adjusted the algorithm to use the new definitions.

index.bs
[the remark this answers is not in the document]
alexnj: Updated with the latter one. PTAL.

index.bs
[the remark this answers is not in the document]
nondebug: Acknowledged, {{octet}}s is fine if it doesn't recognize the plural

index.bs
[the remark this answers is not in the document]
nondebug: This seems like an issue with the base spec, filed #632. Linking the reference seems fine to me.

index.bs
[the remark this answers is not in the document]
nondebug: The current description with bluetooth.BluetoothManufacturerData LGTM

nondebug (approved):
lgtm

index.bs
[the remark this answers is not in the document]
alexnj: Thanks @nondebug!

alexnj:
@reillyeon let me know if this is good to merge.

index.bs
```diff
@@ -4850,11 +4876,31 @@ The <a>default allowlist</a> for this feature is <code>["self"]</code>.
...
 Issue: CDDL snippetes use the "text" type instead of
 "browsingContext.BrowsingContext" to allow indepedent programmatic
 processing of CDDL snippets. Currently, other modules cannot be
 referenced.
 
+## Definitions ## {#bluetooth-bidi-definitions}
+
+<pre class="cddl remote-cddl local-cddl">
+bluetooth.BluetoothServiceUuid = text;
+bluetooth.BluetoothManufacturerData = { key: uint, data: tstr };
```
reillyeon, line 4901: `data` should be a `bstr` since manufacturer data isn't a human readable string, just a sequence of bytes.

reillyeon (approved):

index.bs
alexnj: Updated.

PR #633, opened 2024-10-07
Change BlutoothManufacturerData.data to base64-encoded tstr

alexnj:
This PR is changing [`BluetoothManufacturerData.data` from being a `bstr`](https://github.com/WebBluetoothCG/web-bluetooth/pull/630#discussion_r1783657473) and defining it as a base64 encoded text representation as a `tstr`. Since BiDi is a protocol over JSON, this seems the correct way to represent bytes per [RFC 8610](https://datatracker.ietf.org/doc/html/rfc8610#appendix-E).

***
<a href="https://pr-preview.s3.amazonaws.com/alexnj/web-bluetooth/pull/633.html" title="Last updated on Oct 8, 2024, 5:48 PM UTC (ee06d56)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/633/5fca8d9...alexnj:ee06d56.html" title="Last updated on Oct 8, 2024, 5:48 PM UTC (ee06d56)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4906,7 +4906,7 @@ bluetooth.BluetoothManufacturerData = { key: uint, data: tstr };
   <dd>is the Company Identifier Code.</dd>
 
   <dt><code>data</code></dt>
-  <dd>is the base64 encoded string representing the <a spec="infra">list</a> of {{octet}}s.</dd>
+  <dd>is the [=forgiving-base64 encoded=] result of the manufacturer data [=byte sequence=].</dd>
 </dl>
 
 ## The bluetooth module ## {#bluetooth-module}
@@ -5017,7 +5017,7 @@ A `bluetooth.ScanRecord` represents data of the advertisement packet sent by a [
   <dd>is an <a>Appearance</a>, one of the values defined by the {{gap.appearance}} characteristic.</dd>
 
   <dt><code>manufacturerData</code></dt>
-  <dd>list of <code>bluetooth.BluetoothManufacturerData</code> that maps {{unsigned short}} Company Identifier Codes to [=lists=] of {{octet}}s.</dd>
+  <dd>list of <code>BluetoothManufacturerData</code> that maps {{unsigned short}} Company Identifier Codes to [=forgiving-base64 encoded=] result of manufacturer data [=byte sequences=].</dd>
 </dl>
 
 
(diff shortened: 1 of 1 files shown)


reillyeon:
It seems weird to push how to encode binary data into individual specs. Can we get some guidance from WebDriver BiDi folks on the correct type for this type of argument?

alexnj:
BiDi spec has [one example](https://w3c.github.io/webdriver-bidi/#type-network-BytesValue) that encodes binary data, and that seems to refer to [forgiving-base64 encode](https://infra.spec.whatwg.org/#forgiving-base64-decode). Not sure if the algorithm would be considered lossy given the code point filtering employed, for transporting `ManufacturerData`. 

/cc @OrKoN

OrKoN:
Yes, using base64 and a text string is currently recommended for binary data if the size is relatively small and can be inline in a protocol message.

OrKoN (approved):

OrKoN:
To clarify, I think referencing the [base64 algorithm](https://infra.spec.whatwg.org/#forgiving-base64) in the infra spec is sufficient as we do not have a central place to handle binary data in the WebDriver BiDi spec. Also from the client and implementation perspective the format is command/event-specific and we do not expect to change/specify it centrally. If we have a better way in the future, we can re-visit but so far I know know of plans to have a better way to deal with large binary files but I do not think a solution for that would be applicable here since CDP also implements this data as base64-encoded strings.

index.bs
```diff
@@ -4898,9 +4898,17 @@ referenced.
...
-bluetooth.BluetoothManufacturerData = { key: uint, data: bstr };
+bluetooth.BluetoothManufacturerData = { key: uint, data: tstr };
 </pre>
 
+<dl>
+  <dt><code>key</code></dt>
+  <dd>is the Company Identifier Code.</dd>
+
+  <dt><code>data</code></dt>
+  <dd>is the [=forgiving-base64 encoded=] result of the manufacturer data [=byte sequence=].</dd>
```
reillyeon, line 4909: It feels weird to me to talk about using the forgiving-base64 encode algorithm to encode this data since the browser isn't what is doing the encoding, it is the external script. I think it would make more sense to say that this is "the manufacturer data byte sequence, base64 encoded" and then specify in the algorithm steps which process this command that we use "forgiving-base64 decode" to interpret this externally-provided value.

reillyeon (changes requested):

index.bs
alexnj: Updated. PTAL.

reillyeon (approved):

PR #646, opened 2025-02-21
Update bluetooth.simulateAdapter command

chengweih001:
Update `bluetooth.simulateAdapter` command to support creating, updating, and removing adapter.

***
<a href="https://pr-preview.s3.amazonaws.com/chengweih001/web-bluetooth/pull/646.html" title="Last updated on Feb 21, 2025, 10:23 PM UTC (6d9537f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/646/d2d4d06...chengweih001:6d9537f.html" title="Last updated on Feb 21, 2025, 10:23 PM UTC (6d9537f)">Diff</a>

PR #648, opened 2025-03-05
New bluetooth.disableSimulation command for Web Bluetooth automation

chengweih001:
This pull request moves `type` == `remove` from `bluetooth.simulateAdapter` to a new command `bluetooth.disableSimulation`. The goal is to make it clear that the effect of `bluetooth.simulateAdapter` would end up with a simulated adapter, while `bluetooth.disableSimulation` would remove the simulation entirely.

***
<a href="https://pr-preview.s3.amazonaws.com/chengweih001/web-bluetooth/pull/648.html" title="Last updated on Mar 11, 2025, 1:40 AM UTC (fcdb0ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/648/22c5b49...chengweih001:fcdb0ee.html" title="Last updated on Mar 11, 2025, 1:40 AM UTC (fcdb0ee)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5196,9 +5196,8 @@ bluetooth.simulateAdapter = (
 
 bluetooth.SimulateAdapterParameters = {
    context: text,
-   type: "create" / "update",
    ? leSupported: bool,
-   ? state: "absent" / "powered-off" / "powered-on"
+   state: "absent" / "powered-off" / "powered-on"
 }
 </pre>
 
@@ -5208,22 +5207,18 @@ The [=remote end steps=] with command parameters |params| are:
 1. Let |contextId| be |params|["context"].
 1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
 1. If |navigable| is not a [=navigable/top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
-1. If |params|[`"type"`] is `create`, run the following steps:
+1. Let |simulatedBluetoothAdapter| be |navigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, run the following steps:
     1. If |params|[`"leSupported"`] does not [=map/exist=], return [=error=] with [=error code=] [=invalid argument=].
-    1. If |params|[`"state"`] does not [=map/exist=], return [=error=] with [=error code=] [=invalid argument=].
     1. Let |simulatedBluetoothAdapter| be a new [=simulated Bluetooth adapter=].
     1. Set |simulatedBluetoothAdapter|'s <a>LE supported state</a> to |params|[`"leSupported"`].
     1. Set |simulatedBluetoothAdapter|'s <a>adapter state</a> to |params|[`"state"`].
     1. Set |navigable|'s <a>simulated Bluetooth adapter</a> to |simulatedBluetoothAdapter|.
     1. Return [=success=] with data `null`.
-1. If |params|[`"type"`] is `update`, run the following steps:
+1. If |simulatedBluetoothAdapter| is not empty, run the following steps:
     1. If |params|[`"leSupported"`] [=map/exists=], return [=error=] with [=error code=] [=invalid argument=].
-    1. If |params|[`"state"`] does not [=map/exist=], return [=error=] with [=error code=] [=invalid argument=].
-    1. Let |simulatedBluetoothAdapter| be |navigable|'s <a>simulated Bluetooth adapter</a>.
-    1. If |simulatedBluetoothAdapter| is empty, return [=error=] with [=error code=] [=invalid argument=].
     1. Set |simulatedBluetoothAdapter|'s <a>adapter state</a> to |params|[`"state"`].
     1. Return [=success=] with data `null`.
-1. Return [=error=] with [=error code=] [=invalid argument=].
 
 </div>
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -5196,7 +5196,7 @@ bluetooth.simulateAdapter = (
 
 bluetooth.SimulateAdapterParameters = {
    context: text,
-   type: "create" / "update" / "remove",
+   type: "create" / "update",
```
reillyeon, line 5199: I think `type` still needs to be made optional (with a default of `"create"`) so this doesn't break compatibility.

reillyeon (changes requested):

index.bs
chengweih001: I wonder if we need to consider backward compatibility here? Since the major user is a few test cases in [Web Bluetooth BIDI](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/bluetooth/bidi/), which have already been broken when https://github.com/WebBluetoothCG/web-bluetooth/pull/646 was merged (Chrome Bidi driver has bot to crawl spec change hence Chrome Bidi driver [updated](https://github.com/GoogleChromeLabs/chromium-bidi/pull/3140) and broke tests). 

So I thought we just move forward with without considering backward compatibility.

reillyeon: Realistically, yes, there probably isn't that much code out there which uses the Web Bluetooth BiDi interface and so we can probably change it in incompatible ways, fix our tests and move on, but we probably shouldn't.

chengweih001: Understood. I will update the patch to make the `type` field optional.

reillyeon: If the `type` field is optional we probably don't need to add a separate `disableSimulation` command but I'd appreciate feedback from @sadym-chromium on that aspect of WebDriver BiDi API design.

sadym-chromium: Maybe I lack some context here, but I don't think we don't need to distinguish "create" and "update". Instead, we can have 2 commands: `bluetooth.simulateAdapter` (which creates or updates the adapter, depending on the state) and `bluetooth.disableSimulation`.

reillyeon: True, we could remove the `type` field entirely and treat create or update the simulated depending on whether it is already present.

chengweih001: Thanks Reilly and Maksim for the suggestion. I updated the patch to remove `type`.

index.bs
```diff
@@ -5196,39 +5196,29 @@ bluetooth.simulateAdapter = (
...
 The [=remote end steps=] with command parameters |params| are:
 
-1. Let |contextId| be params["context"].
+1. Let |contextId| be |params|["context"].
 1. Let |navigable| be the result of [=trying=] to [=get a navigable=] with |contextId|.
 1. If |navigable| is not a [=navigable/top-level traversable=], return [=error=] with [=error code=] [=invalid argument=].
-1. If |params|[`"type"`] is `create`, run the following steps:
+1. Let |simulatedBluetoothAdapter| be |navigable|'s <a>simulated Bluetooth adapter</a>.
+1. If |simulatedBluetoothAdapter| is empty, run the following steps:
     1. If |params|[`"leSupported"`] does not [=map/exist=], return [=error=] with [=error code=] [=invalid argument=].
```
reillyeon, line 5212: For compatibility with the API before #646 this needs to assume a default value of `true` for the `leSupported` parameter.

reillyeon (changes requested):

index.bs

reillyeon (approved):

chengweih001:
I made additional changes (https://github.com/WebBluetoothCG/web-bluetooth/pull/648/commits/5a800e1f256d8009e0b168fef557be3276433ab2, https://github.com/WebBluetoothCG/web-bluetooth/pull/648/commits/fcdb0ee1e04e620f50a20e57846fa8f1788d01a7)  to fix the following error:

```
 [INFO] Build bikeshed document "index.bs" (common)…
    $ bikeshed --die-on=warning spec "index.bs" "index.bs.built.html" 
      LINE 1352:49: Multiple possible 'allowed to use' dfn refs.
      Arbitrarily chose https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use
      To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
      spec:html; type:dfn; text:allowed to use
      spec:private-aggregation-api; type:dfn; text:allowed to use
      [=allowed to use=]
      LINE 1442:25: Multiple possible 'allowed to use' dfn refs.
      Arbitrarily chose https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use
      To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
      spec:html; type:dfn; text:allowed to use
      spec:private-aggregation-api; type:dfn; text:allowed to use
      [=allowed to use=]
      LINE 2009:49: Multiple possible 'allowed to use' dfn refs.
      Arbitrarily chose https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use
      To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
      spec:html; type:dfn; text:allowed to use
      spec:private-aggregation-api; type:dfn; text:allowed to use
      [=allowed to use=]
       ✘  Did not generate, due to errors exceeding the allowed error level.
    Command `bikeshed --die-on=warning spec "index.bs" "index.bs.built.html" ` failed with exit code: 2.
```
Is it okay to piggyback the fix in this PR?

sadym-chromium (approved):

reillyeon (approved):

PR #650, opened 2025-03-30
bluetooth.gattConnectionAttempted and bluetooth.simulateGattConnectionResponse for Web Bluetooth automation

chengweih001:
Add new WebDriver Bidi command and event for Web Bluetooth automation:
- bluetooth.gattConnectionAttempted: Event upon GATT connection attempt.
- bluetooth.simulateGattConnectionResponse: For simulating the response for a GATT connection attempt.

***
<a href="https://pr-preview.s3.amazonaws.com/chengweih001/web-bluetooth/pull/650.html" title="Last updated on Apr 16, 2025, 4:42 AM UTC (34b4996)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WebBluetoothCG/web-bluetooth/650/c845e58...chengweih001:34b4996.html" title="Last updated on Apr 16, 2025, 4:42 AM UTC (34b4996)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3144,10 +3144,12 @@ method, when invoked, MUST perform the following steps:
             [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
         1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `null`, wait for
             it to be non-`null`.
-        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is not `0`, do the
+        1. Let |response| be [=this=].{{[[automatedGATTConnectionResponse]]}}.
+        1. Set [=this=].{{[[automatedGATTConnectionResponse]]}} to `null`.
+        1. If |response| is not `0`, do the
             following sub-steps:
             1. Remove |promise| from [=this=].{{[[activeAlgorithms]]}}.
-            1. [=queue a global task=] on the [=Bluetooth task source=] given
+            1. [=Queue a global task=] on the [=Bluetooth task source=] given
                 |global| to [=reject=] |promise| with a "{{NetworkError}}"
                 {{DOMException}} and abort these steps.
     1. Otherwise, run the following steps:
@@ -5527,8 +5529,8 @@ To <dfn>trigger a prompt updated event</dfn> given a [=navigable=] |navigable|, 
 1. Set [=map of navigables to device prompts=][|navigableId|] to |prompt|.
 1. Let |params| be a [=map=] matching the <code>bluetooth.RequestDevicePromptUpdatedParameters</code> production with the <code>context</code> field set to |navigableId|, the <code>prompt</code> field set to |promptId|, and the <code>devices</code> field set to |serialized devices|.
 1. Let |body| be a [=map=] matching the <code>bluetooth.RequestDevicePromptUpdated</code> production, with the <code>params</code> field set to |params|.
-1. Let |related navigables| be a [=/set=] containing |navigable|.
-1. For each |session| in the [=set of sessions for which an event is enabled=] given "<code>bluetooth.requestDevicePromptUpdated</code>" and |related navigables|:
+1. Let |relatedNavigables| be a [=/set=] containing |navigable|.
+1. For each |session| in the [=set of sessions for which an event is enabled=] given "<code>bluetooth.requestDevicePromptUpdated</code>" and |relatedNavigables|:
     1. [=Emit an event=] with |session| and |body|.
 
 </div>
@@ -5552,12 +5554,12 @@ To <dfn>trigger a gatt connection attempted event</dfn> given a [=navigable=] |n
 
 1. Let |navigableId| be |navigable|'s [=navigable id=].
 1. Let |params| be a [=map=] matching the <code>bluetooth.GattConnectionAttemptedParameters</code> production with the
-    <code>context</code> field set to |navigableId|, the <code>address</code> field set to |device|.{{[[representedDevice]]}}'s address.
+    <code>context</code> field set to |navigableId| and the <code>address</code> field set to |device|.{{[[representedDevice]]}}'s address.
 1. Let |body| be a [=map=] matching the <code>bluetooth.GattConnectionAttempted</code> production, with the
     <code>params</code> field set to |params|.
-1. Let |related navigables| be a [=/set=] containing |navigable|.
+1. Let |relatedNavigables| be a [=/set=] containing |navigable|.
 1. For each |session| in the [=set of sessions for which an event is enabled=] given
-    "<code>bluetooth.gattEventGenerated</code>" and |related navigables|:
+    "<code>bluetooth.gattEventGenerated</code>" and |relatedNavigables|:
     1. [=Emit an event=] with |session| and |body|.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -110,6 +110,7 @@ spec: WEBDRIVER; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
         text: error; url: dfn-error
         text: local end; url: dfn-local-ends
+        text: invalid element state; url: dfn-invalid-element-state
 
 </pre>
 <pre class="link-defaults">
@@ -3113,7 +3114,7 @@ slots</a> described in the following table:
   </tr>
   <tr>
     <td><dfn>\[[automatedGATTConnectionResponse]]</dfn></td>
-    <td><code>null</code></td>
+    <td><code>"not-expected"</code></td>
     <td>
       The simulated GATT connection response code for a GATT connection attempt.
     </td>
@@ -3142,10 +3143,12 @@ method, when invoked, MUST perform the following steps:
         is not empty, run the following steps:
         1. [=Trigger a gatt connection attempted event=] given |global|'s
             [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
-        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `null`, wait for
-            it to be non-`null`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"not-expected"`,
+            set it to `"expected"`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"expected"`,
+            wait for it to be changed.
         1. Let |response| be [=this=].{{[[automatedGATTConnectionResponse]]}}.
-        1. Set [=this=].{{[[automatedGATTConnectionResponse]]}} to `null`.
+        1. Set [=this=].{{[[automatedGATTConnectionResponse]]}} to `"not-expected"`.
         1. If |response| is not `0`, do the
             following sub-steps:
             1. Remove |promise| from [=this=].{{[[activeAlgorithms]]}}.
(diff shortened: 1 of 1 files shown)


chengweih001:
Hi Reilly,

Could you please take a look at this patch to see if this pattern is on the right track to support these commands? I tried to leverage `upon fulfillment of promise` to make the special steps for simulation limited to only around the text for performing GATT connection and service discovery.

Additionally, for the following PR for supporting add/removing services/characteristics/descriptors, I plan to have those  simulation commands create promises in [Bluetooth.[[attributeInstanceMap]]](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetooth-attributeinstancemap-slot), so that they will align with what I proposed in this PR.

index.bs
```diff
@@ -5429,6 +5457,77 @@ A [=local end=] could simulate a device advertisement by sending the following m
...
+   type: "connection" / "discovery",
+   code: uint
+}
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateGattResponse">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be params["context"].
```
reillyeon, line 5480:
```suggestion
1. Let |contextId| be |params|["context"].
```

index.bs
```diff
@@ -5429,6 +5457,77 @@ A [=local end=] could simulate a device advertisement by sending the following m
...
+1. If |simulatedBluetoothAdapter| is empty, return [=error=] with [=error code=] [=invalid argument=].
+1. Let |deviceMapping| be |simulatedBluetoothAdapter|'s <a>simulated Bluetooth device mapping</a>.
+1. If |deviceMapping|[|deviceAddress|] [=map/exists=], let |simulatedDevice| be |deviceMapping|[|deviceAddress|].
+    Otherwise, return [=error=] with [=error code=] [=invalid argument=].
+1. Let |simulatedDeviceInstance| be the result of <a>get the <code>BluetoothDevice</code> representing</a>
+    |simulatedDevice| inside |navigable|'s <a>active window</a>'s <a spec=HTML>associated <code>Navigator</code></a>'s
+    [=associated Bluetooth=].
+1. Let |gatt| be |simulatedDeviceInstance|.{{[[gatt]]}}.
+1. For each |promise| in |gatt|.{{[[activeAlgorithms]]}}, [=resolve=] |promise| with |params| and remove |promise|
+    from |gatt|.{{[[activeAlgorithms]]}}.
```
reillyeon, line 5493: This isn't resolving the promise created in "query the Bluetooth cache" with the same type as that algorithm does. We probably need to add a new slot that holds a "waiting for the automation script to provide a response" promise which algorithms like that can wait for.

There's a similar problem for handling `connect()`. I would fix that example first as I'm not sure an explicit "discovery" event is exactly how we want to model this for automation.

reillyeon (changes requested):

index.bs

index.bs
chengweih001: Thank you for the discussion today. I revised the patch to use a new internal slot for holding pending automation response and remove service discovery event.

index.bs
```diff
@@ -3129,6 +3137,27 @@ method, when invoked, MUST perform the following steps:
...
 1. Add |promise| to [=this=].{{[[activeAlgorithms]]}}.
+1. If |global|'s [=Window/navigable=]'s
+    [=navigable/top-level traversable=]'s <a>simulated Bluetooth adapter</a> is
+    not empty, run the following steps:
+    1. [=Trigger a gatt connection attempted event=] given |global|'s [=Window/navigable=]
+        and [=this=].{{BluetoothRemoteGATTServer/device}}.
+    1. Let |pendingAutomationPromise| be a [=a new promise=].
+    1. Add |pendingAutomationPromise| to
+        [=this=].{{[[pendingAutomationGATTConnectionResponse]]}}.
+    1. <a>Upon fulfillment</a> of |pendingAutomationPromise| with |responseCode|,
```
reillyeon, line 3148: Checking with other standards experts it sounds like we can simplify this to not use promises but instead "wait for" the value in the internal slot to change. So, we would give it an initial value like "null" and then "wait for this.[[automatedGattConnectionRespons]] to be non-null".

reillyeon (changes requested):

index.bs
```diff
@@ -5429,6 +5458,57 @@ A [=local end=] could simulate a device advertisement by sending the following m
 </pre>
 </div>
 
+#### The bluetooth.simulateGattConnectionResponse Command #### {#bluetooth-simulategattconnectionresponse-command}
+
+<pre highlight="cddl" class="cddl remote-cddl local-cddl">
+bluetooth.SimulateGattConnectionResponse = (
```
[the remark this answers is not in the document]
chengweih001: Done.

index.bs
```diff
@@ -5429,6 +5458,57 @@ A [=local end=] could simulate a device advertisement by sending the following m
...
+   address: text,
+   code: uint
+}
+
+</pre>
+
+<div algorithm="remote end steps for bluetooth.simulateGattConnectionResponse">
+The [=remote end steps=] with command parameters |params| are:
+
+1. Let |contextId| be |params|["context"].
```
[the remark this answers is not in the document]

index.bs
chengweih001: Thanks for the suggestion. I revised the patch.

So I assume that if we are doing this way, for `bluetooth.simulateDisconnectionEvent`, we would also reset this `[[automatedGattConnectionResponse]]` slot in order to keep the state right for future connection attempt?

reillyeon: I think the easier way to model that is reset `[[automatedGattConnectionResponse]]` at the completion of the `connect()` call. If `connect()` is called again before `disconnect()` we already have logic to handle that.

index.bs
```diff
@@ -3130,43 +3137,57 @@ method, when invoked, MUST perform the following steps:
...
+        [=navigable/top-level traversable=]'s <a>simulated Bluetooth adapter</a>
+        is not empty, run the following steps:
+        1. [=Trigger a gatt connection attempted event=] given |global|'s
+            [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `null`, wait for
+            it to be non-`null`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is not `0`, do the
+            following sub-steps:
+            1. Remove |promise| from [=this=].{{[[activeAlgorithms]]}}.
+            1. [=queue a global task=] on the [=Bluetooth task source=] given
```
reillyeon, line 3150:
```suggestion
            1. [=Queue a global task=] on the [=Bluetooth task source=] given
```

index.bs
```diff
@@ -5453,14 +5525,43 @@ To <dfn>trigger a prompt updated event</dfn> given a [=navigable=] |navigable|,
...
+   address: text
+}
+</pre>
+
+<div algorithm="remote end event trigger for bluetooth.GattConnectionAttempted">
+To <dfn>trigger a gatt connection attempted event</dfn> given a [=navigable=] |navigable| and a {{BluetoothDevice}} |device|:
+
+1. Let |navigableId| be |navigable|'s [=navigable id=].
+1. Let |params| be a [=map=] matching the <code>bluetooth.GattConnectionAttemptedParameters</code> production with the
+    <code>context</code> field set to |navigableId|, the <code>address</code> field set to |device|.{{[[representedDevice]]}}'s address.
```
reillyeon, line 5555:
```suggestion
    <code>context</code> field set to |navigableId| and the <code>address</code> field set to |device|.{{[[representedDevice]]}}'s address.
```

index.bs
```diff
@@ -5453,14 +5525,43 @@ To <dfn>trigger a prompt updated event</dfn> given a [=navigable=] |navigable|,
...
+1. Let |body| be a [=map=] matching the <code>bluetooth.GattConnectionAttempted</code> production, with the
+    <code>params</code> field set to |params|.
+1. Let |related navigables| be a [=/set=] containing |navigable|.
```
reillyeon, line 5558:
```suggestion
1. Let |relatedNavigables| be a [=/set=] containing |navigable|.
```

reillyeon (changes requested):

index.bs

index.bs

index.bs

index.bs

index.bs
```diff
@@ -3130,43 +3137,59 @@ method, when invoked, MUST perform the following steps:
...
-            specification allows the UA to opportunistically create one on
-            platforms where that's possible. This may cause a user-visible
-            pairing dialog to appear when a connection is created, instead of
-            when a restricted characteristic is accessed.
+    1. If |global|'s [=Window/navigable=]'s
+        [=navigable/top-level traversable=]'s <a>simulated Bluetooth adapter</a>
+        is not empty, run the following steps:
+        1. [=Trigger a gatt connection attempted event=] given |global|'s
+            [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `null`, wait for
```
reillyeon, line 3145: Sorry, I realize we should probably be even stricter in how these events are ordered: We should set `[[automatedGATTConnectionResponse]]` to some sentinel value like `"expected"` here, and wait for it to be set by the automation command. Then after consuming the response we set it back to an initial value like `"not-expected"` so that the automation command can return an error if the script tries to provide a response before the browser is ready for it.

reillyeon (changes requested):

index.bs

index.bs
```diff
@@ -3142,10 +3143,12 @@ method, when invoked, MUST perform the following steps:
         is not empty, run the following steps:
         1. [=Trigger a gatt connection attempted event=] given |global|'s
             [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
-        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `null`, wait for
-            it to be non-`null`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"not-expected"`,
+            set it to `"expected"`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"expected"`,
```
chengweih001, line 3148: I put this way in consideration of `connect` being called multiple times while waiting for simulation to set a response. The second `connect` would just see it is ` expected and wait there.

index.bs
```diff
@@ -3130,43 +3138,61 @@ method, when invoked, MUST perform the following steps:
...
-            when a restricted characteristic is accessed.
+    1. If |global|'s [=Window/navigable=]'s
+        [=navigable/top-level traversable=]'s <a>simulated Bluetooth adapter</a>
+        is not empty, run the following steps:
+        1. [=Trigger a gatt connection attempted event=] given |global|'s
+            [=Window/navigable=] and [=this=].{{BluetoothRemoteGATTServer/device}}.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"not-expected"`,
+            set it to `"expected"`.
+        1. If [=this=].{{[[automatedGATTConnectionResponse]]}} is `"expected"`,
+            wait for it to be changed.
```
reillyeon, line 3149:
```suggestion
            wait for it to change.
```

index.bs
```diff
@@ -3130,43 +3138,61 @@ method, when invoked, MUST perform the following steps:
 1. Let |promise| be [=a new promise=].
```
reillyeon, line 3138: Check if this.connected is true and if so immediately return a resolved promise.

reillyeon (approved):

index.bs

